PR #34, opened 2016-06-16
Add option to share file for android and iOS

superandrew213:
This allows to share and open files with available apps. My use case was to send to Instagram.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,15 @@ Share Social , Sending Simple Data to Other Apps
   - Add `using Cl.Json.RNShare;` to the usings at the top of the file
   - Add `new RNSharePackage()` to the `List<IReactPackage>` returned by the `Packages` method
 
+## Options
+Share.open() accepts the following options.
+
+Option | Description
+------ | -----------
+title | Android only
+share_text | Text to share
+share_URL | Url to share
+share_File | Path of file to share or open (android & iOS only)
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Option | Description
 title | Android only
 share_text | Text to share
 share_URL | Url to share
-share_File | Path of file to share or open (android & iOS only)
+share_file | Path of file to share or open (android & iOS only)
 
 ## Usage
 
--- a/android/src/main/java/cl/json/RNShareModule.java
+++ b/android/src/main/java/cl/json/RNShareModule.java
@@ -1,5 +1,9 @@
 package cl.json;
 
+import java.io.File;
+
+import android.net.Uri;
+import android.webkit.MimeTypeMap;
 import android.content.Intent;
 import android.content.ActivityNotFoundException;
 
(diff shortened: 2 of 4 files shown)


bcalik:
Is it possible to pass share_File param as url to image? Thanks.

superandrew213:
@bcalik yes. Share_file param is the absolute url of the image without ```file:``` prefix. On iOS you can pass this as the uri. On android I think you have to add the ```file:``` prefix.

bcalik:
@superandrew213 Thank you. I mean an image which is not stored on the device, a http url.

superandrew213:
Then it won't work. It's only meant to work for internally stored files. For urls you would need to use share_URL.

Other option would be to download and store the file using react-native-fs and then use share_file.

superandrew213:
@EstebanFuentealba what do you think? Can we add this option?

vcalvello:
Hey guys! any news on this?

superandrew213:
This package has changed quite but since I forked it. I will have a look at it again but it will take me a bit. Feel free to use my fork in the meantime.

adamkdean:
@EstebanFuentealba this was closed on 22 Sep – could you elaborate why? Was Instagram integrated?

vinayan3:
If the URL is like `file://something.something` then there is an intention of sharing a document. Instagram shows up in the list.  It looks like this:
http://i.imgur.com/UxT7lQj.png

airman00:
Does this work for video? The Menu appears with all the options, except Instagram. I've checked the More submenu too looking for IG. I can share to FB and Photo Library just fine.

If I save the video to Photos on the device, I can actually share to Instagram from the Photos app.

I've tried with .mp4 and .mov format for the video. the video is 13 seconds long.

can someone try it with video? Curious if its an error on my end or a bug in the library.

Thanks!

iamakimmer:
Will this PR be merged in? or what is needed to get this up and running?

danrice01:
@airman00 Did you get this working with video in the end?

airman00:
@danrice01 
The trick is to make sure you don't set a title or subject.

Here's an example that worked for me
`  shareVideo(){
      var finalVideoPath = 'file://' + RNFS.DocumentDirectoryPath + '/finalVideo.mp4'

      let shareOptions = {
        //  title: "React Native",  // don't put a title
          message: "Hola mundo",
          url: finalVideoPath,
          type: "video/mp4",
       //   subject: "Share Link" //  don't put a subject
        };
    Share.open(shareOptions);

    `

danrice01:
@airman00 Thank you. Working :)

sibelius:
any changes we get this in?

nidzola-nikolaz:
any progress on this?

ataillefer:
Guys, this PR couldn't be merged as it stands because as @superandrew213 said the fork was made more than 1 year ago! Since then, `react-native-share` has integrated file sharing which is the purpose of this PR.

PR #62, opened 2016-09-15
fixed local file sharing

logileifs:
I couldn't get local file sharing on android to work so I added a functionality to read in the local file as base64 and then share. I will admit that this solution is a bit hacky but I'm creating the pull request anyway as is and will be happy to fix if you have any comments.

### DIFF
--- a/android/src/main/java/cl/json/ShareFile.java
+++ b/android/src/main/java/cl/json/ShareFile.java
@@ -9,19 +9,18 @@ import android.os.Environment;
 import android.provider.MediaStore;
 import android.util.Base64;
 import android.webkit.MimeTypeMap;
+import android.util.Log;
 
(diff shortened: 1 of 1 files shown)


EstebanFuentealba:
Hi thanks . I'll check

ataillefer:
@logileifs I'm interested in the issue you were facing, what exactly was the error when trying to share a local file on Android? Was it a `NullPointerException` in `ShareFile#getRealPathFromURI` by any chance (I can see that you have commented its code)?

I have been facing a similar issue with local files for which the mime type wasn't correctly detected, created pull request #63 to solve it. It might interest you and I would be glad to have your feedback in case it seems useful to you :smile:

enieber:
this pr is very old and has conflicts

PR #63, opened 2016-09-19
Allow to pass file mime type as an option

ataillefer:
On Android (6.0), for some reason that I can't explain, when the mime type of a local file cannot be guessed from its URL I get a `NullPointerException` at https://github.com/EstebanFuentealba/react-native-share/blob/92aaa3f87e88377cdc74c6b272e5b37543473c4a/android/src/main/java/cl/json/ShareFile.java#L90

This happens for instance with files like `myImage.JPG` (case issue), `myFile` (no extension) or `myMarkdown.md` (unknown extension), for which `ShareFile#getMimeType` returns `null`, thus the call to `ShareFile#getRealPathFromURI` in `ShareFile#isLocalFile`.

Note that I download and store files with [react-native-fetch-blob](https://github.com/wkh237/react-native-fetch-blob) configured to use the Android DownloadManager. An example of storage location is `file:///data/user/0/com.android.providers.downloads/cache/myPDFFile.pdf`.

The purpose of this pull request is to allow to explicitely pass the mime type in the options when sharing a local file:

    Share.open({
      url: fileURL,
      type: fileMimeType,
    }

This definitley solves my issue and I believe it can be useful when knowing the mime type of a local file at the moment we need to share it.

PS: looking at #62, I wonder if @logileifs wasn't facing the same issue, so might be linked.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ Supported options:
 | Name  | Type     | Description |
 | :---- | :------: | :--- |
 | url | string   | URL you want to share (you can share a base64 file url only in iOS & Android ) |
+| type | string   | File mime type (optional) |
 | message | string   |  |
 | title | string   |  (optional) |
 | subject | string   | (optional) |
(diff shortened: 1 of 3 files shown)


logileifs:
@ataillefer sounds like the issue I was having (the path to the file is made up like my path) although in my case it didn't have anything to do with the mime type but the share component was expecting the file to be somewhere public like media store.

logileifs:
and although I fixed the location in the sharing component the problem still remains since the app I was sharing to didn't have access to any files in my app's local storage

PR #68, opened 2016-09-30
Add promise rejection on iOS if app not installed

asciiman:
Adding promise rejection for whatsapp and generic apps on ios.  Don't need to do for email or google plus since they don't rely on specific apps being installed.


EstebanFuentealba:
Gracias 👍

PR #83, opened 2016-11-25
add web sharing fallback for facebook and twitter on iOS

micabe:


boytainnt:
@micabe hi Micabe, i have a problem when sharing via twitter and facebook. I use your code and it work but i have a problem:
screen shot 2017-01-09 at 9 26 34 am

PR #91, opened 2017-01-02
Remove height from action sheet when not visible

iRoachie:
The action sheet shouldn't have a height or appear in the layout when it's not visible. The problem was that even though the sheet wasn't visible, it was taking up space on the layout and blocking other buttons.

This fix hides the action completely when it's not visible.

### DIFF
--- a/ios/EmailShare.h
+++ b/ios/EmailShare.h
@@ -8,11 +8,11 @@
 
 
 #import <UIKit/UIKit.h>
-#import "RCTConvert.h"
-#import "RCTBridge.h"
-#import "RCTUIManager.h"
-#import "RCTLog.h"
-#import "RCTUtils.h"
+#import <React/RCTConvert.h>
+#import <React/RCTBridge.h>
+#import <React/RCTUIManager.h>
+#import <React/RCTLog.h>
+#import <React/RCTUtils.h>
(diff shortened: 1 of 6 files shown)


stoufa88:
I'm also facing this problem. For example, I have a Toolbar with icon (Back arrow) being hidden by the Sheet overlay.

index.js
```diff
@@ -85,7 +86,10 @@ class ShareSheet extends React.Component {
   render(){
     return (
       <Overlay visible={this.props.visible} {...this.props}>
-        <View style={styles.actionSheetContainer}>
+        <View style={[styles.actionSheetContainer, 
+        {
+          height: this.props.visible ? Dimensions.get('window').height : 0
+        }]}>
```
wasedaigo: Instead of "removing" the height here, you can fix Overlay. Which seems to be a better fix to me. What do you think?
https://github.com/wasedaigo/react-native-share/commit/398aa738c8d3fc7de4f43f1c158dc3866313b1c2

iRoachie: Ohh well that's even cleaner. I like that

wasedaigo: Thanks for the response. I made  a PR #101 
I think you can close this PR.

PR #95, opened 2017-01-08
Fixed compatibility with React Native 0.40.0

dcgoss:
See https://github.com/facebook/react-native/releases/tag/v0.40.0
“breaking changes”

Fixes #92

PR #99, opened 2017-01-14
Fixed compatibility with React Native 0.40.0

### LINKED ISSUE #92: doesn't work with React Native v0.40
duplicate interface definition for class 'RCTBridge'
screen shot 2017-01-05 at 12 37 41 am

techieshark:
See https://github.com/facebook/react-native/releases/tag/v0.40.0
“breaking changes”

--- 

Update: I've taken Pull Request #95 from @dcgoss and just ran `git rebase -i master` to merge the two commits into one, so the commit history can be a bit more clear. I've made no other change.

I've also tested this on a new project before and after this change, and can verify I see the compile problem in #92 and can verify that this fixes that compile problem (fixes #92).

### DIFF
--- a/ios/EmailShare.h
+++ b/ios/EmailShare.h
@@ -8,11 +8,11 @@
 
 
 #import <UIKit/UIKit.h>
-#import "RCTConvert.h"
-#import "RCTBridge.h"
-#import "RCTUIManager.h"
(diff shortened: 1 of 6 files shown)


iRoachie:
Fixed in #95

PR #121, opened 2017-04-28
Fix broken link to WhatsApp in the iTunes store.

scovetta:
The original link isn't resolving, so changed to a non-locale-specific iTunes URL.

While I was there, changed HTTP -> HTTPS.

### DIFF
--- a/ios/WhatsAppShare.m
+++ b/ios/WhatsAppShare.m
@@ -28,7 +28,7 @@
             successCallback(@[]);
         } else {
           // Cannot open whatsapp
-          NSString *stringURL = @"http://itunes.apple.com/en/app/whatsapp-messenger/id310633997";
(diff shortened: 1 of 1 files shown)


sibelius:
is this a problem in android as well?

scovetta:
@sibelius I think Android is fine -- the only [reference](https://github.com/EstebanFuentealba/react-native-share/blob/master/android/src/main/java/cl/json/social/WhatsAppShare.java#L14) I can find is to `market://details?id=com.whatsapp`, which looks correct as per [this doc](https://developer.android.com/distribute/marketing-tools/linking-to-google-play.html#android-app).

jgcmarins:
@sibelius can we merge this?
thanks to @wasedaigo and @scovetta

nicholasess:
@sibelius Do you found someone problem with the PR?

PR #138, opened 2017-07-06
Support sending an email to a predefined email address

mohamed-salah:
support sending an email to predefined email address
for android and ios


enieber:
need documentation how to use this props to merge

PR #149, opened 2017-08-12
Simplify JavaScript code in components

fxfactorial:
1) Removed unused code
2) Removed class when it was not necessary


sibelius:
I think we need setup prettier in this repo

and I prefer to keep the name of classes

PR #155, opened 2017-08-30
Fix breaking change for RN 0.47

alfonsogoberjr:
Removed reference to createJSModules() which breaks Android builds. 

See [facebook/react-native@ce6fb33](https://github.com/facebook/react-native/commit/ce6fb337a146e6f261f2afb564aa19363774a7a8)


sibelius:
fixed on master 

tks

PR #169, opened 2017-10-04
Fix check with Java way

jgcmarins:
Sorry, I've forgot that `&&` Java operator only works with booleans.


sibelius:
makes sense

tks

PR #175, opened 2017-10-09
Fixed SLComposeViewController for iOS 11

samermurad:
SLComposeViewController.isAvailableForServiceType is unneeded and always returns false, iOS handles these errors (case no facebook/twitter user is available.
this is true for older iOS versions as well, tested on 9,9.3,10.3.1 and 11.
once trying to load the SLComposeViewController, if no social account is present, the user will simply get a message around the lines of "No facebook/twitter account is configured, this can be done from the settings".
SLComposeViewController.isAvailableForServiceType is always returning false and I needed the share to open the facebook/twitter popup inside the app, this works without the isAvailableForServiceType.

I expanded the method GenericShare.singleShare to include shouldOpenInWeb flag, just in case someone wants to open the share in the browser.

### DIFF
--- a/ios/GenericShare.h
+++ b/ios/GenericShare.h
@@ -50,6 +50,6 @@
 #import "React/RCTUtils.h"   // Required when used as a Pod in a Swift project
 #endif
 @interface GenericShare : NSObject <RCTBridgeModule>
-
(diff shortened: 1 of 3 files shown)


rohitgoyal:
Looks like this could be the reason image sharing is failing for Facebook in iOS 11.

PR #181, opened 2017-10-19
Add file provider

ygorbarboza:
Solve `android.os.FileUriExposedException: file.jpg exposed beyond app through ClipData.Item.getUri()` exception on targetSdk 24+

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-native-share",
   "description": "Social Share, Sending Simple Data to Other Apps",
-  "version": "999.0.26",
+  "version": "999.0.27",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/EstebanFuentealba/react-native-share.git"
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-native-share",
   "description": "Social Share, Sending Simple Data to Other Apps",
-  "version": "199999.0.23",
+  "version": "999.0.23",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/EstebanFuentealba/react-native-share.git"
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-native-share",
   "description": "Social Share, Sending Simple Data to Other Apps",
-  "version": "999.0.28",
+  "version": "1.0.23",
   "repository": {
     "type": "git",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "react-native-share",
   "description": "Social Share, Sending Simple Data to Other Apps",
-  "version": "1.0.23",
+  "version": "999.0.28",
```
jgcmarins: why is that?

ygorbarboza:
@jgcmarins sorry about that. Actually this PR was not suppose to be created now. This change is only relevante when React Native uses a new targetSdkVersion. For now there is no need for this

deliveryblueprints:
Thanks for this, works great. One thing to note to get this working in your project is to implement the ShareApplication.java in your MainApplication.java.
Then override the getFileProviderAuthority like so..
`  @Override
    public String getFileProviderAuthority() {
        return BuildConfig.APPLICATION_ID + ".fileprovider";
    }
`

This will properly populate the xml with your applicationId, i.e. com.myapporg.myappname.fileprovider

Thanks again!

ER1011:
@deliveryblueprints, what exactly did you put in the MainApplication.java and where does the @Override part go? I've tried in multiple ways but can't get it to work. Thank you

ygorbarboza:
@ER1011 Just do as said here, make your application extends `ShareApplication` and create function`getFileProviderAuthority` with the same name as the provider-> android:authorities on Manifest

ER1011:
@ygorbarboza the link you posted seems to be broken

ygorbarboza:
@ER1011 sorry about that
https://developer.android.com/training/secure-file-sharing/setup-sharing.html

jgcmarins:
probably solves https://github.com/EstebanFuentealba/react-native-share/issues/185

ygorbarboza:
@jgcmarins Yup

ygorbarboza:
This can be merged, the problem is that we gonna have to bump the compileSdkVersion and the buildTools

jgcmarins:
@EstebanFuentealba and @sibelius can we merge this PR?

bitshadow:
Thanks @ygorbarboza for PR. We really needed this in our app.

bitshadow:
@sibelius Looks like this is working for us can you release the new package in NPM if possible. Thanks.

PR #197, opened 2018-01-17
Add ShareApplication explanation on README.md

ygorbarboza:


README.md
```diff
@@ -45,6 +45,24 @@ Share Social , Sending Simple Data to Other Apps
     ```
       compile project(':react-native-share')
     ```
+5. Follow this [guide](https://developer.android.com/training/secure-file-sharing/setup-sharing.html)
+6. Make your `Application` class implements `ShareApplication`
+ - Make `getFileProviderAuthority` function return the `android:authorities` that was added on AndroidManifest file
+
+Example: 
+```
+class MyApplication extends Application implements ShareApplication, ReactApplication {
```
bitshadow: How do we import ShareApplication?

README.md
```diff
@@ -49,6 +49,20 @@ Share Social , Sending Simple Data to Other Apps
...
 
+Example: 
+```
+class MyApplication extends Application implements ShareApplication, ReactApplication {
+
+...
+
+     @override
+     String getFileProviderAuthority() {
+            return <File provider authority>
```
sibelius: could you provide an example of this?

README.md
ygorbarboza: I added the import on the example

README.md
ygorbarboza: this string should be the same you added in the AndroidManifest. If a add a fixed string on the example can be misleading

sibelius: got it

bitshadow:
Thanks @ygorbarboza its working for me now.

PR #198, opened 2018-01-17
Update README  

bitshadow:
Continuing #197 after making it work in actual app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,20 +46,30 @@ Share Social , Sending Simple Data to Other Apps
       compile project(':react-native-share')
     ```
 5. Follow this [guide](https://developer.android.com/training/secure-file-sharing/setup-sharing.html)
+Example:
+Put this in `AndroidManifest.xml` where `applicationId` is something that you have defined in `android/app/build.gradle`.
+  <applicaiton>
+    <provider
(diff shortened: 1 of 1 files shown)


mikehardy:
@jgcmarins Possibly worth noting that if you use rn-fetch-blob, and you follow the README as is now after this merge - Android builds will have a collision with the AndroidManifest provider element. In my case, with rn-fetch-blob installed I did not have to add the paths XML or the provider entry in the manifest, I just had to override the Java method getFileProviderAuthority in MainApplication. I don't know if it's useful or not to have cross-references to other projects (like rn-fetch-blob) from this repo's README but if it bites other people, that duplication of configuration between the two repos is the reason

jgcmarins:
I'll reimplement example repo and reproduce all README steps.
I'll also try to add an example with rn-fetch-blob to understand what's the catch.
Thanks for referencing this PR.

mikehardy:
FWIW, here is what I went with in the react-native-update-apk repo that has collision possibilities with rn-fetch-blob (frequently used with react-native-share) so that I could avoid the support costs of debugging Android FileProviders:

https://github.com/mikehardy/react-native-update-apk#manual-steps-for-android

Issue #200, opened 2018-01-20
com.share01.MainApplication cannot be cast to cl.json.ShareApplication


monkingame:
I created a test react-native project,but got error:
com.share01.MainApplication cannot be cast to cl.json.ShareApplication

steps:
1)react-native init share01
2)yarn add react-native-share
3)react-native link
4)copy the sample codes to replace index.js
5)react-native run-android and got error

react-native environment:
Environment:
  OS: macOS High Sierra 10.13.2
  Node: 9.3.0
  Yarn: 1.3.2
  npm: 5.6.0
  Watchman: 4.9.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: 16.2.0 => 16.2.0
  react-native: 0.52.0 => 0.52.0

so,what's the error?

Thanks!

![35183311-e198f35e-fe1e-11e7-80e6-f2e10a5a19d7](https://user-images.githubusercontent.com/4568610/35188703-9bebadcc-fe75-11e7-916a-e0af96d64d84.jpg)

giturra:
Same issues here

maxocg:
Same issues here

jgcmarins:
@ygorbarboza maybe this is related to your fix? What do you think?

sibelius:
Follow README instructions

monkingame:
@sibelius @jgcmarins Thanks !
I think the lastest version(v1.0.24) maybe have some bugs or I do not know how to config.
Here are my steps to test(I tested 3 version: v1.0.21 v1.0.23 v1.0.24):

1)react-native init TestShare
the test project will use v1.0.21
and two other project:TestShare23 TestShare24
2)yarn
install node_moudles
)using android studio to open TestShare/android project(for building)
)open android emulator
)react-native run-android
(All above steps sure no problems)
)yarn add react-native-share@1.0.21
here I tested 3 versions in different projects.
version test:1.0.21/1.0.23/1.0.24
)react-native link
)react-native run-android
For rebuilding
)copy sample codes to override index.js
)press r twice to reload app in emulator
)click 3 sample buttons
[Simple Share Image base64]:
v1.0.21 v1.0.23 OK
v1.0.24 got error:
com.testshare24.MainApplication cannot be cast to cl.json.ShareApplication
[Simple Share]:
All versions OK
[Share UI Component]:
All versions OK
[...]

PR #228, opened 2018-03-22
Add missing Android documentation

tribou:
As requested in #200

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,12 +77,13 @@ Share Social , Sending Simple Data to Other Apps
       </paths>
     ```
 
-6. Make your `Application` class implements `ShareApplication`
-  - Make `getFileProviderAuthority` function return the `android:authorities`
-    that was added on AndroidManifest file
-  - Example: Replace `com.example.yourappidhere` with the `applicationId` that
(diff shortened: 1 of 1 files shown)


kylegoetz:
Hi. Was reading through the changes in hopes of getting this plugin to work, and saw a modification to the function `getFileProviderAuthority` to modify. Where is this? I've disabled the .gitignore and run a search for this term and get zero results. I've installed this and done `react-native link`. Currently running into the same error it seems everyone else is: `com.reactnativetest.MainApplication cannot be cast to cl.json.ShareApplication

tribou:
Hi @kylegoetz, you're right. Reading back over that part of the documentation, that part may not be clear enough.

You actually have to add the `getFileProviderAuthority` method to your `MainApplication` class (it didn't exist in my project beforehand). Here's an example gist of what the `android/app/src/main/java/com/example/myapp/MainApplication.java` file will look like after adding it:

https://gist.github.com/tribou/00f06ca302ba974b02f3830297674a77#file-mainapplication-java-L52-L55

kylegoetz:
@tribou thank you. I got it working. @EstebanFuentealba thanks for the library.

PR #246, opened 2018-04-26
Allow specifying reject/resolve for user cancel action for sharing on iOS

### LINKED ISSUE #245: Allow param on `open` function to determine whether to reject if user cancels share
Feature request:

Currently, the `open` method rejects if the user clicks "cancel" in the share menu. Would it be possible/desirable to pass an optional param to not reject on user cancel? Something like this:

```
Share.open({
... options,
failOnCancel: false, // would default to true, since that's current behavior
}).catch((err) => {
handleError(err); // we'd no longer get the "User did not share" error here since failOnCancel = false
});
```

Happy to look into it and if successful, submit a PR, just wanted to know if that's something people are interested/ok with adding.

sam1463:
New optional param for `Share.open` method to cause promise to resolve on user cancel action, rather than rejecting.

This only affects iOS. On android this module already does not reject if user cancels, although there appear to be a couple of requests for supporting that functionality.

Resolves #245

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ Supported options:
 | title | string   |  (optional) |
 | subject | string   | (optional) |
 | excludedActivityTypes | string   | (optional) |
-| failOnCancel | boolean | (defaults to true) on iOS, specifies whether promise should reject if user cancels share dialog (optional) |
+| failOnCancel | boolean | (defaults to true) On iOS, specifies whether promise should reject if user cancels share dialog (optional) |
 
 #### shareSingle(options) (in iOS & Android)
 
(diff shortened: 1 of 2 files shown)


nicholasess:
@sam1463 please, can you resolve the conflicting?

sam1463:
@nicholasess resolved

README.md
```diff
@@ -142,7 +142,9 @@ Supported options:
 | title | string   |  (optional) |
 | subject | string   | (optional) |
 | excludedActivityTypes | string   | (optional) |
-| showAppsToView | boolean | (optional) only android| 
+| failOnCancel | boolean | (defaults to true) On iOS, specifies whether promise should reject if user cancels share dialog (optional) |
+| showAppsToView | boolean | (optional) only android|
+>>>>>>> upstream/master
```
jgcmarins: remove this, please

sam1463: 👍 not sure how I missed that before. It's removed now.

jgcmarins: thanks 🎉

sam1463:
Anything I can do to help get this merged? Is it just a matter of needing more reviewers?

jgcmarins:
thanks @sam1463

PR #287, opened 2018-06-25
update README with Troubleshooting section how to share PDF files for iOS

BigPun86:


BigPun86:
@jgcmarins nice, i am really happy to hear that!!

PR #291, opened 2018-06-26
fix issue #290

enieber:


jgcmarins:
Thanks, but solved here: https://github.com/react-native-community/react-native-share/pull/292

Issue #248, opened 2018-04-27
I am able to send base64 image only for first time Android


HiteshGunwant:
Hey guys ,
I am able to share base 64 image only for one time for next time in Android I am calling this fun it just displaying message as "Test",And when i kill app next time it again working same as follow
pls let me know solution URJENT!!
Thanks,
  let shareOptions = {
                        "url": 'data:image/png;base64,'+ base64Str,
                        "social": "twitter",
                        "message":"Test",
                        }
                        console.log("shareOptionsshareOptions.base64Image,",shareOptions)
                  
                        Share.shareSingle(shareOptions);

jgcmarins:
@HiteshGunwant could you check if #244 solved this?

evilchis94:
@jgcmarins as @HiteshGunwant I was unable to share base64 image nor downloaded image, it works only for the first time. My Android version is 5.0.1 and targetSdkVersion is 25, I'm asking for WRITE_EXTERNAL_STORAGE before sharing base64 and it always return true.



            let hasPermission = false;
            try {
                const check = await PermissionsAndroid.check(PermissionsAndroid.PERMISSIONS.WRITE_EXTERNAL_STORAGE);

                console.log('permiso de escritura ', check)

                if (check) {
                    hasPermission = true;
                }
                else {
                    const granted = await PermissionsAndroid.request(PermissionsAndroid.PERMISSIONS.WRITE_EXTERNAL_STORAGE,
                        {
                            'title': 'Cool Photo App Camera Permission',
                            'message': 'Cool Photo App needs access to your camera ' +
                                       'so you can take awesome pictures.'
                        }
                    )
    
                    if (granted === PermissionsAndroid.RESULTS.GRANTED) {
                        hasPermission = true;
                    }
                }
            } catch (err) {
                console.warn(err)
            }

            if (hasPermission) {
                RNFetchBlob
                .fetch('GET', image)
                .then(res => {

                    let url = `data:image/${extension};base64,${res.base64()}`

                        Share.shareSingle({
                                url,
                                title: 'Share on Facebook',
                                social: 'facebook'
                            })
                            .then(async (res) => {
                                // await RNFetchBlob.fs.unlink(res.path());
                            })
                            .catch(async (error) => {
                                console.log(error)
                            })

                })
            }

This is the message

![img_20180522_135640_575](https://user-images.githubusercontent.com/10981327/40384068-260a736a-5dc8-11e8-80eb-9771b4415df1.JPG)

[...]

PR #299, opened 2018-07-04
Android new intent initialize on every request and Constants expose

kkbhav:
A new instance of share class will be created every time shareSingle method is called.  
This solves issues like #248  

Also I have exposed social constants for easier access on JS side

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,12 +111,6 @@ Share Social , Sending Simple Data to Other Apps
 
     }
     ```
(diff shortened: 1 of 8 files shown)


kkbhav:
Hey, I cannot merge this request, as it is asking for commits to be signed. I tried to sign all the commits, but git created new commits with the signed request. Can you please check, if we can do something about this

PR #309, opened 2018-07-24
little improvement over too many "def"

yeomann:
Android def cleanup: add a ternary based function with fallback, just a little improved version rather then having many explicit def for each value.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -8,7 +8,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:+'
+        classpath 'com.android.tools.build:gradle:3.1.2'
     }
 }
 
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -14,19 +14,18 @@ buildscript {
 
 apply plugin: 'com.android.library'
 
-def DEFAULT_COMPILE_SDK_VERSION = 26
-def DEFAULT_BUILD_TOOLS_VERSION = "26.0.2"
-def DEFAULT_TARGET_SDK_VERSION = 26
-def DEFAULT_GOOGLE_PLAY_SERVICES_VERSION = "12.0.1"
-def DEFAULT_SUPPORT_LIBRARY_VERSION = "27.1.0"
+def safeExtGet(prop, fallback) {
+    rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
+}
 
 android {
-    compileSdkVersion rootProject.hasProperty('compileSdkVersion') ? rootProject.compileSdkVersion : DEFAULT_COMPILE_SDK_VERSION
-    buildToolsVersion rootProject.hasProperty('buildToolsVersion') ? rootProject.buildToolsVersion : DEFAULT_BUILD_TOOLS_VERSION
+    compileSdkVersion safeExtGet('compileSdkVersion', 26)
+    buildToolsVersion safeExtGet('buildToolsVersion', '26.0.3')
+
 
     defaultConfig {
-        minSdkVersion 16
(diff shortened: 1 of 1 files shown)


jgcmarins:
```
FAILURE: Build failed with an exception.

* Where:
Build file '/root/app/android/build.gradle' line: 15

* What went wrong:
A problem occurred evaluating root project 'android'.
> Failed to apply plugin [id 'com.android.library']
   > Minimum supported Gradle version is 4.9-rc-1. Current version is 4.8. If using the gradle wrapper, try editing the distributionUrl in /root/app/android/gradle/wrapper/gradle-wrapper.properties to gradle-4.9-rc-1-all.zip

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 25s
Exited with code 1
```

yeomann:
@jgcmarins I have inspected a little bit and what I understood so far the above error you are getting is because we Gradle wrapper we are targeting 4.8 version like so `distributionUrl=https\://services.gradle.org/distributions/gradle-4.8-all.zip` but on the build file for Gradle there is version specified i.e 
```
dependencies {
        classpath 'com.android.tools.build:gradle:+'
    }
```

after more testing I came to realise that 

> This version of the Android plugin requires the following: 
Gradle 4.4 or higher. To learn more, read the section about updating Gradle.
Build Tools 27.0.3 or higher. Keep in mind, you no longer need to specify a version for the build tools using the android.buildToolsVersion property—the plugin uses the minimum required version by default 

Ref link: https://developer.android.com/studio/releases/gradle-plugin

As project is using latest Gradle version already then it would be good to target SDK 27 since along required by Android 

After upgrading SDK version, I got another error i.e

```
 What went wrong:
A problem occurred configuring root project 'android'.
> Could not resolve all artifacts for configuration ':classpath'.
   > Could not resolve com.android.tools.build:builder:3.3.0-alpha03.
     Required by:
         project : > com.android.tools.build:gradle:3.3.0-alpha03
      > Could not resolve com.android.tools.build:builder:3.3.0-alpha03.
         > Could not get resource 'https://jcenter.bintray.com/com/android/tools/build/builder/3.3.0-alpha03/builder-3.3.0-alpha03.pom'.
            > Could not GET 'https://jcenter.bintray.com/com/android/tools/build/builder/3.3.0-alpha03/builder-3.3.0-alpha03.pom'.
               > Read timed out
   > Could not resolve com.android.tools.analytics-library:shared:26.3.0-alpha03.
     Required by:
         project : > com.android.tools.build:gradle:3.3.0-alpha03
```

which clearly shows that its 404 when gradle tried to download latest Plugin version as shown above in error `3.3.0-alpha03`

I will change that `gradle:+` to a stable version, because as Android advice use to use specific version not a dynamic 

> Caution: You should not use dynamic dependencies in version numbers, such as 'com.android.tools.build:gradle:2.+'. Using this feature can cause unexpected version updates and difficulty resolving version differences.

so let me test that real quick as well.

yeomann:
ok as you can @jgcmarins I was correct on my assumption. Hence, Test is passed now after I sent the Gradle plugin version because dynamic is advised from Android itself.

There's a room for one more improvement since we are already using latest Gradle and targeting SDK, we can safely remove "compile" and replaced it with the "implementation" As this is actually a new requirement for Android and we must do it because otherwise we will see this error

**"Configuration 'compile' is obsolete and has been replaced with 'implementation' or 'Api'"**
  
more about the Old and new configuration please see this https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration?utm_source=android-studio#new_configurations
at the heading of "**Use the new dependency configurations**"

I will submit a separate PR for that.

yeomann:
pleasure @jgcmarins

jgcmarins:
@yeomann I've invited you to the repo.

yeomann:
hey @jgcmarins  cheers for that. I will try to give my best input!

PR #312, opened 2018-07-25
Update obsolete compile to implementation

yeomann:
ref to closed PR https://github.com/react-native-community/react-native-share/pull/311

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -50,6 +50,6 @@ repositories {
 }
 
 dependencies {
-    implementation "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
+    compileOnly "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
     implementation "com.android.support:appcompat-v7:${safeExtGet('supportLibVersion', '27.1.0')}"
 }
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -50,6 +50,6 @@ repositories {
 }
 
 dependencies {
-    compile "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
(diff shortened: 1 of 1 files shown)


yeomann:
alright, test seems good as well. @jgcmarins and Hi @sibelius :)

android/build.gradle
```diff
@@ -50,6 +50,6 @@ repositories {
 }
 
 dependencies {
-    compile "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
-    compile "com.android.support:appcompat-v7:${safeExtGet('supportLibVersion', '27.1.0')}"
+    implementation "com.facebook.react:react-native:${safeExtGet('reactNativeVersion', '+')}"
```
sibelius: In fact this should be `compileOnly` https://github.com/react-native-community/react-native-camera/blob/master/android/build.gradle#L48

check react-native-camera

yeomann: Hi, I don't know the purpose why they are using in camera plugin "compileOnly" I always check for the official documentation, I believe maybe CompileOnly can work as well. Normally the different is that "implementation" tells Gradle that the dependency is available  at runtime only. where "compileOnly" as from the name says is called at the time of compiling the app saying that this might not depend upon app. This also applies that CompileOnly are the Android libraries which are not depending obviously on React-native.  

Read more here https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration?utm_source=android-studio#new_configurations

yeomann: actually you are right, let me change and test it.

yeomann: Nice catch @sibelius test seems good as well

jgcmarins:
Great one
`compileOnly` is the right way to do that.
Thanks.

PR #345, opened 2018-08-17
[Android] Fix SingleShareIntent to always call super.open

### LINKED ISSUE #337: Share.shareSingle(shareOptions) does not pass data to Facebook app
I want to share image to Facebook. With `Share.open(shareOptions)` it works but when I am trying to use `Share.shareSingle(shareOptions)` it does not pass data to Facebook app. It only opens the Facebook's new status page with no data on it.

christianbach:
Fixes #337


christianbach:
@jgcmarins thanks for the review! Would it be possible to merge it to master?

PR #346, opened 2018-08-24
rn link react-native-share

lfkwtz:
avoid side effects by specifying pkg


jgcmarins:
this will also avoid lots of issues, thanks!!!

Issue #344, opened 2018-08-17
Using Share.shareSingle() on Android gives back an unknown error


IronTony:
### Steps to reproduce
1. I am using
```
const shareOptions = {
      url: picture.uri,
      social: Share.Social.INSTAGRAM,
};
Share.shareSingle(shareOptions).then(() => console.log('SUCCESS')).catch(err => console.log('ERROR: ', err));
```
2. The Instagram app seems to start to open, but while it is opened the screen is blank and I have back an error: "An unknown error occured"

### Expected behaviour
It has to open the Instagram app as it happens in the iOS app.

### Actual behaviour
It seems that Instagram app will open, but the screen stays blank and I have a toast back with "An unknown error occured"

### Environment
- **React Native version**: 0.55.4
- **React Native platform + platform version**: Android 7.1.2

### react-native-share
**Version**: npm version

If you need more infos please tell me.

jgcmarins:
looks like this is a problem with your link: https://github.com/react-native-community/react-native-share/search?q=An+unknown+error+occured&unscoped_q=An+unknown+error+occured

andreipop-reea:
I have the same problem when trying to Share.shareSingle() on Instagram. It seems to be an issue with the Instagram service, because other works properly.
I tested on Android 7/8 with the latest version of react-native-share v1.1.1

vforvasile:
Hey while Share.open works, I could not make shareSingle to Instagram to work. Any suggestions? My setups is same as asker's.

nicubarbaros:
Same here.
Have anyone tried the previous version?

IronTony:
@jgcmarins hi, here is a small (fast written) project that shows the problems: https://github.com/IronTony/RNShareExample
Hope it can help.

jgcmarins:
@ddereszewski awesome.
Could you send a PR fixing it?

ddereszewski:
The fix has been merged to master, can you update and check if works for you.

jgcmarins:
it is already on latest release: https://github.com/react-native-community/react-native-share/releases/tag/v1.1.2

PR #365, opened 2018-09-12
Fix singleShare for social Media

ddereszewski:
We discover an issue with SingleShare on Android. Details explained here #344


ddereszewski:
Please allow some additional testing before merging.

jgcmarins:
@ddereszewski what do you suggest?

ddereszewski:
The original commit didn't work on Android 6. Now It's good go!

ddereszewski:
We don't use ReactNativeShare on IOS as we use direct Linking, so I can't tell from experience. I looked into the IOS implementation and it's similar to what we have on RN. 

Here the issue is android specific, I wouldn't expect problems on IOS, but I'm happy to help in future if something occurs and have a bit of time.

PR #371, opened 2018-09-25
Add pinterest in ANDROID

deathart:
Hello,

Here is the system to share on pinterest on android
I'm not the skills for IOs, so don't hesitate to take over the PR:)

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,9 +115,9 @@ Share Social , Sending Simple Data to Other Apps
 
     }
     ```
-  
+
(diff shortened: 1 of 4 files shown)


deathart:
It is feasible I think, but I have no knowledge of IOS so I can't do it :s

jgcmarins:
@deathart that's ok. Are you already using this on production?

deathart:
Yes I use it in production in a personal project

PR #373, opened 2018-10-01
Android dependencies

mikehardy:
I'm not super familiar with react-native-share yet, but I'm pretty familiar with android development and I thought maybe you guys could use a dependency refresh?

This quiets all the warnings you'd see with Android Studio current (3.2 now) by upgrading gradle, gradle wrapper scripts, repository definitions, and library versions.


example/package.json
```diff
@@ -9,7 +9,7 @@
   "dependencies": {
     "react": "16.3.1",
     "react-native": "0.55.4",
-    "react-native-share": "git+https://github.com/kkbhav/react-native-share.git"
```
jgcmarins: thanks

example/android/app/build.gradle
```diff
@@ -137,10 +136,10 @@ android {
...
 
 dependencies {
-    compile project(':react-native-share')
-    compile fileTree(dir: "libs", include: ["*.jar"])
-    compile "com.android.support:appcompat-v7:23.0.1"
-    compile "com.facebook.react:react-native:+"  // From node_modules
+    implementation project(':react-native-share')
+    implementation fileTree(dir: "libs", include: ["*.jar"])
+    implementation "com.android.support:appcompat-v7:28.0.0"
+    implementation "com.facebook.react:react-native:+"  // From node_modules
```
sibelius: this should be compileOnly

mikehardy: I'm looking into this - I'm more familiar with apps (which use libraries via implementation) then library development such as here, where compileOnly might make sense. I think you're right on first glance but am verifying with docs right now.

mikehardy: ...and found the answer right after. Yes, compileOnly is better as client apps will have this declared. I'll alter this

mikehardy: ...then contradicted myself with testing. I just pushed two changes after verifying what's necessary for the example to work out of the box. This comment was on the example app, so it must have these as implementation as it needs them at runtime - in the library (the parent of the example) compileOnly could work, but this doesn't run unless they are implementation.

The other change was that the relative path I attempted in the package.json of the example - a reference to '..' for the necessary react-native-share dependency resulted in the bundler becoming confused about where to resolve things. The original reference to some other github repo for the react-native-share dep didn't seem right, so I put in the official repo and I successfully run the example out of the box now

example/package.json
mikehardy: if only it actually worked! relative path resulted in bundler confusion - pushed a change that at least references the official repo though

mikehardy:
I just re-pushed with cleaner commit history, and current deps (android moves about as fast as react-native land it seems...)

PR #432, opened 2019-01-22
Sharing internal storage files

a-koka:
Added instructions to use files from internal storage on Android

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,10 @@ If you want to share Whatsapp and Mailto, you should write `LSApplicationQueries
     ```xml
       <?xml version="1.0" encoding="utf-8"?>
       <paths xmlns:android="http://schemas.android.com/apk/res/android">
+          <-- For files shared using base64 URLs. Required WRITE_EXTERNAL_STORAGE permission -->
           <external-path name="myexternalimages" path="Download/" />
+          <-- For files in internal storage under myInternalDir -->
+          <files-path path="myInternalDir/" name="myInternalImages" />
       </paths>
     ```
 
--- a/README.md
+++ b/README.md
@@ -95,9 +95,9 @@ If you want to share Whatsapp and Mailto, you should write `LSApplicationQueries
     ```xml
       <?xml version="1.0" encoding="utf-8"?>
       <paths xmlns:android="http://schemas.android.com/apk/res/android">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -95,7 +95,10 @@ If you want to share Whatsapp and Mailto, you should write `LSApplicationQueries
     ```xml
       <paths xmlns:android="http://schemas.android.com/apk/res/android">
+          <-- For files shared using base64 URLs. Required WRITE_EXTERNAL_STORAGE permission -->
           <external-path name="myexternalimages" path="Download/" />
+          <-- For files in internal storage under myInternalDir -->
```
mikehardy: This looks good in general, but I think the last bit needed to make sure the "a-ha!" light bulb goes on for people would be to say:

`<-- For files in internal storage "myInternalDir" (replace with your path) -->`

jgcmarins: nice catch

mikehardy:
Also, I looked at CircleCI when I saw this, and I had some thoughts:

1. The CI fail has nothing to do with this PR. It should proceed independent of CI status
1. The CI fail is related to Android tools + API license acceptance
1. The license acceptance is done in a brittle way in Circle (vs Travis, which I'm more familiar with) and is not "really" in our control. I use the scare-quotes there because there are two workarounds you can do to have accepted licenses - you can either accept them locally, then copy the magic license acceptance blobs from your local machine to your repo, and then in CI copy them from the repo to the correct spot, or you can run some acceptance scripts. Or (final alternative) you can wait for Circle CI to update their Docker image.

I honestly can't recommend doing the magic blob thing, it's terrible and requires intervention in a distasteful way every update. The scripts might be a better way - they would still require intervention every Android tooling update but at least it isn't distasteful, just possibly brittle. The best thing to do might be to ignore Circle CI and wait for them to fix but I hate advocating to ignore CI, it is only for the reality of not wanting to invest time in learning Circle CI YAML scripting myself (I already know Travis, I'm good, thanks ;-) )

I must admit I'm a little shocked Circle CI doesn't keep their Docker images super up to date, much less with a lag like they have now. These tools have been out for a couple months, and they are a tool company?

README.md
a-koka: Thanks @mikehardy. I pushed the suggested changes.

mikehardy:
I think this is great, it should help people in the future. +1 karma 😁

PR #442, opened 2019-01-30
SingleShare forceDialog option

ddereszewski:
This pull request resolves Android workflow issue with some social medias singleShare.

### The way it works now 
User share an image to instagram and see this 
img_0350
### Problem
1. user clicks Always and share to Instagram Feed
2. User shares again, and the app opens Instagram Feed by default 

User is not being able to share to Instagram Story anymore, as Android remember decision.

### Proposed solution 
On Android I suggest option **forceDialog**, which disables buttons Always / Just once. Instead, it always forces to choose the option. It looks as attached below.

img_8474

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,7 @@ Supported options:
 | title | string   |  (optional) |
 | subject | string   | (optional) |
 | social | string   | supported social apps: [List](#static-values-for-social)  |
+| forceDialog | boolean | (optional) only android. Avoid showing dialog with buttons Just Once / Always. Useful for Instagram to always ask user if share as Story or Feed |
 
 ***NOTE: If both `message` and `url` are provided `url` will be concatenated to the end of `message` to form the body of the message. If only one is provided it will be used***
 
--- a/android/src/main/java/cl/json/social/SingleShareIntent.java
+++ b/android/src/main/java/cl/json/social/SingleShareIntent.java
@@ -55,7 +55,13 @@ public abstract class SingleShareIntent extends ShareIntent {
         super.open(options);
     }
     protected void openIntentChooser() throws ActivityNotFoundException {
-        this.getIntent().setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
-        this.reactContext.startActivity(this.getIntent());
+        if(this.options.hasKey("forceDialog") && this.options.getBoolean("forceDialog")){
+             Intent chooser = Intent.createChooser(this.getIntent(), this.chooserTitle);
+             chooser.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+             this.reactContext.startActivity(chooser);
+        }else{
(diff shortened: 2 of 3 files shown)


ddereszewski:
I think you need to change CricleCi to use build tools 28

I see in the settings 
`ANDROID_BUILD_TOOLS_VERSION: "27.0.2"`

mikehardy:
Either I am not understanding CircleCI (a definite possibility), or this is not fixable without using some different Android SDK license strategy - https://github.com/react-native-community/react-native-share/pull/432#issuecomment-456540961

I just cloned this repo cleanly and ran the same commands circleci is trying to do, for what it's worth, and master checks out. 

I also added this branch as a remote, fetched it and ran the commands, and it checks out.

Finally, I checked out the branch from #432 and checked it, and it also seems fine

I don't intend to provide personal CI services but I hate seeing PRs held up :-), hopefully the CircleCI thing resolves (either by a PR from someone or CircleCI changing things, or ???) soon but these seem okay...

mikehardy:
> check this: [react-native-community/react-native-camera:circle.yml@`master`](https://github.com/react-native-community/react-native-camera/blob/master/circle.yml)


Yeah - react-native-camera are using old build tools - 26.0.2, and I think the CircleCI image they use has pre-accepted those, so everything is fine. But react-native-share has moved Android dependencies forward (as it should / as should be possible)

Following that CircleCI link (thanks!) it appears CircleCI has altered the whole way they do android images and we need to alter the image like so: https://discuss.circleci.com/t/licences-for-android-sdk-not-accepted/27830/5

...which is to say, no I was not understanding CircleCI correctly / deeply-enough, but if we make this change to circle.yml it should work

mikehardy:
@ddereszewski if you try that different image name, maybe CI will go green? (with API28 I suppose)

ddereszewski:
I will solve it. Need some time

jgcmarins:
> I will solve it. Need some time

Great!
I can take a look sooner.

ddereszewski:
@jgcmarins @mikehardy Ready to go! 
After merge new release is much appreciated

mikehardy:
Hey - great work - I looked at the commit and though CircleCI is all new to me, it looks like it lines up with the docs and obviously it works ;-). If the other PRs are rebased onto this it should help clear the queue.

jgcmarins:
Well, I guess we can do this today.
I see no breaking changing on diff: https://github.com/react-native-community/react-native-share/compare/v1.1.3...master
Could you help me with review?

jgcmarins:
I also need some help to write a Changelog
cc @muhzi4u

mikehardy:
> I see no breaking changing on diff: [v1.1.3...master](https://github.com/react-native-community/react-native-share/compare/v1.1.3...master)
> Could you help me with review?

I didn't see anything breaking either - I just read through it in case an extra set of eyes could help. My changes related to de-linting made it a little noisy and there shouldn't be any semantic changes with them, but I'll keep an ear tuned for related issues post-release just in case....

PR #453, opened 2019-02-19
perf(deps): update deps, cleanup project, remove old example

jgcmarins:
### missing
- [ ] add updated version of `example`


mikehardy:
If you are thinking of doing the example with rn-fetch-blob (to highlight possible manifest merger issues) I just saw that Android Studio has a "Merged Manifest View" which is gold as it highlights the different manifests that your app and any libraries might be providing, shows the priorities and the final merged output, without having to build anything - https://developer.android.com/studio/build/manifest-merge#inspect_the_merged_manifest_and_find_conflicts

android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-all.zip
```
mikehardy: Just out of curiosity - gradle-4.x-current is 4.10.3 right now, why move back to 4.7?
React-Native itself has moved to 4.10.2 semi-recently - https://github.com/facebook/react-native/tree/master/gradle/wrapper and 4.10.3 seems to work fine for me?

To be clear, I don't think this is a problem exactly and you may have a good reason (I havent' checked through issues), but if there's no reason then staying current seems better

android/build.gradle
```diff
@@ -46,3 +46,8 @@ dependencies {
     //noinspection GradleCompatible
     implementation "com.android.support:appcompat-v7:${safeExtGet('supportLibVersion', '28.0.0')}"
 }
+
+task wrapper(type: Wrapper) {
+    gradleVersion = '4.7'
```
mikehardy: related to gradle versioning comment

android/gradle/wrapper/gradle-wrapper.properties
jgcmarins: sorry, my bad, will fix it

android/build.gradle
jgcmarins: great, thanks

PR #459, opened 2019-02-28
[WIP] Fix SMS Sharing for Android

guerandelv:


index.js
```diff
@@ -147,23 +147,29 @@ class RNShare {
...
-    WHATSAPP: NativeModules.RNShare.WHATSAPP || 'whatsapp',
-    INSTAGRAM: NativeModules.RNShare.INSTAGRAM || 'instagram',
-    INSTAGRAM_STORIES: NativeModules.RNShare.INSTAGRAM_STORIES || 'instagram-stories',
-    GOOGLEPLUS: NativeModules.RNShare.GOOGLEPLUS || 'googleplus',
-    EMAIL: NativeModules.RNShare.EMAIL || 'email',
-    PINTEREST: NativeModules.RNShare.PINTEREST || 'pinterest',
-  };
 
-  static InstagramStories = {
+  static Social = (Platform.OS === 'android')
```
jgcmarins: why do we need this?

guerandelv:
I would not recommend merging my fork right now as I have disabled all ios functionality for react native share. I will come back to it at some point to make sms sharing work seamlessly on android and ios.

ethanyuwang:
Is SMS sharing for Android still broken? I'm getting "TypeError: undefined is not a function (evaluating 'Object.keys(styles)[typeof Symbol === "function" ? Symbol.iterator: "@@iterator"]'()'"

mikehardy:
@ethanyuwang - the module works and there is an example to follow but even with those, when I first used it I got that error. It is terrible because the root cause has nothing to do with the error. I don't have the link to the comment now but basically that's some sort of poor behavior on the Promise library's part that is what you see, and if you patch it, then you get the real error - which if it was like mine was that you weren't handling a Promise correctly with aync/try+await/catch. I believe I was missing the await part and got this. Scan your code and see if you see anything

PR #476, opened 2019-04-18
Fixing some flow types on index.js on isPackageInstalled function

MateusAndrade:
# Overview

Fixing some flowTypes on isPackageInstalled function promise and packageName param.

# Test Plan
Clone the PR, and call the isPackageInstalled function.

### DIFF
--- a/index.js
+++ b/index.js
@@ -162,8 +162,7 @@ class RNShare {
   static InstagramStories = {
     SHARE_BACKGROUND_IMAGE: NativeModules.RNShare.SHARE_BACKGROUND_IMAGE || 'shareBackgroundImage',
     SHARE_STICKER_IMAGE: NativeModules.RNShare.SHARE_STICKER_IMAGE || 'shareStickerImage',
-    SHARE_BACKGROUND_AND_STICKER_IMAGE:
-      NativeModules.RNShare.SHARE_BACKGROUND_AND_STICKER_IMAGE || 'shareBackgroundAndStickerImage',
+    SHARE_BACKGROUND_AND_STICKER_IMAGE: NativeModules.RNShare.SHARE_BACKGROUND_AND_STICKER_IMAGE || 'shareBackgroundAndStickerImage',
   };
 
   static open(options: Options | MultipleOptions): Promise<OpenReturn> {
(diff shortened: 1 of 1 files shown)


mikehardy:
Hi there! Fixing types is awesome, but it's hard for me to say just by looking whether this is valid or not. I mean, developer to developer, this looks obvious - but what if my mind is playing tricks on me?

Could you add a flow check package.json run script (handy example: https://github.com/react-native-community/react-native-device-info/blob/master/package.json#L13) to make sure the result is good, and if it looks like there's a good place to wedget it in the circle CI config make sure it is called there? Somewhere around here https://github.com/react-native-community/react-native-share/blob/master/circle.yml#L18 should work I think - with the flow-bin being a new devDependency as needed I think...

Most importantly (vs the obvious change here) it should make sure it doesn't regress in the future

MateusAndrade:
@mikehardy sure! I just got a doubt, or even a suggestion: is there any contrib guide or something like that? Because i got stuck trying to commit my changes with commitlint. And i dont found any doc explaining how should i commit, etc.

mikehardy:
Alas, there is not. I remember attempting to send a patch to this library was in fact the first time that I was confronted by my total lack of knowledge of all of react-native (with husky, commitlint, changelog formats, etc etc etc). I'm not sure I'd be qualified to write it either but I know @jgcmarins has recently added some documentation (issue templates etc) and he may have set up the package.json originally - maybe he knows?

jgcmarins:
Awesome discussion!
There are few rules here: https://github.com/react-native-community/react-native-share/blob/44ac820e77bc90f331490320a509a52d630272b2/commitlint.config.js
And there is missing a contributing guide explaining how commit lint works.

MateusAndrade:
@jgcmarins would be really amazing if someone write a guide about commit lint, because i needed to use --no-verify to skip the lint process as there is any example about this

jgcmarins:
here you go: https://github.com/react-native-community/react-native-share/issues/477
let's work on this

mikehardy:
I like that this was merged - but I was hoping for the flow check in CI prior to make sure it is good and doesn't regress later. I did not test it locally

jgcmarins:
Agreed, we can add the flow check step on another PR

MateusAndrade:
@mikehardy  i agree with that. I will look at the CI/Flow Check on this weekend =D

MateusAndrade:
#480

PR #480, opened 2019-04-20
Feature/adding flow check ci

MateusAndrade:
# Overview
Adding a flow-check on package.json. Adding flow-definitions for modules like react-native, etc. And it was added a flow-check on CI.

# Test Plan
Run a `yarn flow-check`. it needs to run without any error or warning to ensure that there are no mistaken on type definitions.


MateusAndrade:
I think this would be a better aproach. Um going to see If there is another option about flow types than adding all types

MateusAndrade:
It would be better merge this pr, and when changing to version 1.3.0 add typescript?

MateusAndrade:
@jgcmarins i agree with you. As im a newbie with flow i got this mistake about type-defs. But i found another solution [here](https://github.com/flowtype/flow-for-vscode/issues/215) about setting the type-defs manually.

In this case, since we got react-native and StyleSheet as types, i think using this approach can be ok. What do you think?

flow-defs-libs/react-native.js
```diff
@@ -0,0 +1,7 @@
+declare module 'react-native' {
+  declare module.exports: any
```
jgcmarins: we gonna need better types here

flow-defs-libs/react-native.js
```diff
@@ -0,0 +1,7 @@
...
+}
+
+declare module 'react-native/Libraries/StyleSheet/StyleSheet' {
+  declare module.exports: any
```
jgcmarins: same here

flow-defs-libs/react-native.js
MateusAndrade: Yes, i just write this types because that there are only two being used. But, i see on some issues, and we'have discussed about typescript. 

Would be a better option, release a new version with the master. Then refactor flow to typescript?

jgcmarins: > Would be a better option, release a new version with the master. Then refactor flow to typescript?
Yep, I think so.

You can find me at React Brasil Slack or react-native-community Slack to discuss more about that.

PR #482, opened 2019-04-22
Calling success callback at right time in Android

kkbhav:
# Overview

- In Android, the successCallback or failureCallback for `open` and `shareSingle` methods were called before the actual share dialog disappears which caused a lot of confusion. Also, there was no way to determine the activity which was selected from the chooser dialog. 
For this, I have implemented a BroadcastReceiver which listens to the chooser dialog and notifies the app and ActivityEventListener which listens for the cancel event. This way we will be able to determine if the actual share was initiated or was cancelled midway
- Added `RNSharePathUtil` class to handle multiple types of content URI and building fileProvider Uri

# Test Plan
I have tested this with an internal application in my organisation and it is working smoothly. But I was not able to cover all the application scenarios for `shareSingle`.

# Breaking
Now, success callback reply for Android is similar to that of iOS. Which means anyone checking for the reply as "OK" would encounter an error in their code
Also, I have changed intent flag from **FLAG_ACTIVITY_NEW_TASK** to **FLAG_ACTIVITY_NO_HISTORY**.


jgcmarins:
Wow! This is a huge improvement.
We gonna need few tests. Maybe it is better to have a release before merging it.

Question: why changing from **FLAG_ACTIVITY_NEW_TASK** to **FLAG_ACTIVITY_NO_HISTORY**?

We also gonna need to understand the effects of the breaking change.
@mikehardy comments?

mikehardy:
Yeah - I need to do a little research to think about the activity flag change, but in general this is amazing and should be a huge improvement.

Separately - I never like code on master unreleased, so I'm 100% for a new release as a separate priority, I remember there being some action requested prior but I can't recall what it is (apologies)

mikehardy:
Some thoughts prior to deeper research to really think about it -

- this is a breaking change and a powerful new feature, but there is no demonstration in the example app - could you update that to demonstrate the functionality? you mention you have this working in an internal app so perhaps it's a quick code paste into the example to show it? That will help me greatly in reviewing the change as well, and allow us to point future module consumers to working code for the breaking change upgrade

- this will require documentation as well I believe? I think for broadcast receivers you have to add an AndroidManifest.xml entry don't you?

kkbhav:
@jgcmarins: 
Flag is changed so that we can receive results in onActivityResult, by using **FLAG_ACTIVITY_NEW_TASK** the result is always called before the chooser closes. I replaced it with **FLAG_ACTIVITY_NO_HISTORY** so that no history is maintained for this activity and a new activity is created every time a share happens.
I believe the purpose of **FLAG_ACTIVITY_NEW_TASK** was to create a new activity irrespective of the previous share

@mikehardy 
- I will update the example by tomorrow so that result can be seen properly
- Broadcast receiver is registered when share flow is called and gets removed when flow ends, it does not have a need for registering in Manifest.
- We will have to document the change in result format and also I am thinking of changing reject response to an actual error object rather than a custom object.
- Also, `activityType` is not available pre lollipop(sdk 22), so in that scenario older response "OK" is sent as the `activityType`

jgcmarins:
> I am thinking of changing reject response to an actual error object rather than a custom object.

This could be a good idea.

android/src/main/java/cl/json/social/SingleShareIntent.java
```diff
@@ -54,14 +49,24 @@ public void open(ReadableMap options) throws ActivityNotFoundException {
...
     protected void openIntentChooser() throws ActivityNotFoundException {
-        if(this.options.hasKey("forceDialog") && this.options.getBoolean("forceDialog")){
-             Intent chooser = Intent.createChooser(this.getIntent(), this.chooserTitle);
-             chooser.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
-             this.reactContext.startActivity(chooser);
-        }else{
-             this.getIntent().setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
-             this.reactContext.startActivity(this.getIntent());
+        if (this.options.hasKey("forceDialog") && this.options.getBoolean("forceDialog")) {
+            if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP_MR1) {
```
mikehardy: Should use the static isSupported() method . you have on TargetChosenReceiver to avoid propagating build versions everywhere

android/src/main/java/cl/json/social/ShareIntent.java
```diff
@@ -130,9 +138,17 @@ protected static String urlEncode(String param) {
 
         return extraIntents;
     }
+
     protected void openIntentChooser() throws ActivityNotFoundException {
-        Intent chooser = Intent.createChooser(this.getIntent(), this.chooserTitle);
-        chooser.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+        Intent chooser;
+        IntentSender intentSender = null;
+        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.LOLLIPOP_MR1) {
```
mikehardy: Should use the static isSupported() method . you have on TargetChosenReceiver to avoid propagating build versions everywhere

android/src/main/java/cl/json/social/TargetChosenReceiver.java
```diff
@@ -0,0 +1,91 @@
...
+        failureCallback = failure;
+    }
+
+    @TargetApi(Build.VERSION_CODES.LOLLIPOP_MR1)
+    public static IntentSender getSharingSenderIntent(ReactContext reactContext) {
+        synchronized (LOCK) {
+            if (sTargetChosenReceiveAction == null) {
+                sTargetChosenReceiveAction = reactContext.getCurrentActivity().getPackageName() + "/" + TargetChosenReceiver.class.getName() + "_ACTION";
+            }
+            Context context = reactContext.getCurrentActivity().getApplicationContext();
```
mikehardy: I have seen in other places that the current activity can sometimes go null on you in react-native, unexpectedly. Is there a reason not to use the ReactContext directly? https://github.com/facebook/react-native/issues/18345

Basically if you just need the context you shouldn't have to chain through the activity to get it if I recall correctly

android/src/main/java/cl/json/social/TargetChosenReceiver.java
```diff
@@ -0,0 +1,91 @@
...
+
+import com.facebook.react.bridge.Callback;
+import com.facebook.react.bridge.ReactContext;
+
+/**
+ * Receiver to record the chosen component when sharing an Intent.
+ */
+public class TargetChosenReceiver extends BroadcastReceiver {
+    private static final String EXTRA_RECEIVER_TOKEN = "receiver_token";
+    private static final Object LOCK = new Object();
```
mikehardy: Is this so the app can support multiple shares "in flight" at the same time, and when they come back (possibly to different receivers?) they will resolve correctly for the callbacks?

Is that something that actually can happen? It seems like the app can't have control back from the share dialog until something has been chosen, meaning the callback would happen before the app could be used to share again. If that is correct, then perhaps all the parallel-aware stuff here can fall away? Or is this something that is documented as possible and you actually saw?

kkbhav: This is just to avoid getting a call from anywhere else in the code. If by some mistake any event gets triggered it may lead to false positives. This here takes care of that, so that we can achieve a result as intended.

android/src/main/java/cl/json/social/TargetChosenReceiver.java
kkbhav: Yes, that's my bad. We only need context here. So, reactContext can be used directly. I will make the changes and make a commit regarding this.

kkbhav: I have updated the code to use reactContext as it is.

android/src/main/java/cl/json/social/SingleShareIntent.java
kkbhav: I have updated the code to use isSupported() in place of build versions.

android/src/main/java/cl/json/social/ShareIntent.java
kkbhav: I have updated the code to use isSupported() in place of build versions.

kkbhav:
@mikehardy I have also implemented file provider in the library itself so as to remove the need for manually linking `ShareApplication`. This way we can use file provider for the download use case. Also, if the library user wishes to implement their own provider they can do that by following the given instructions and then their provider will be used to build the `Uri` in place of ours

kkbhav:
Any updates? I have fixed all the issues that were pointed out. 
Is there anything else that remains

mikehardy:
well, it's true that you addressed all the concerns, but the title of this PR is "Calling success callback at right time in Android", and then you altered FileProvider things. So, is this a breaking change, or not? How does this affect the documentation? The example - you gave me a whole lot more to think about at least. If you want a PR to go through, keep them small and don't alter functionality during review except that which is being reviewed is a good guideline I think. If you could separate the fileprovider things into a separate PR then I could focus on the callback work, read through it again, and make sure it's okay for merge. Then the FileProvider could go in after. As it is, I don't have time just now to think through the FileProvider stuff, so no update there, sorry

kkbhav:
@mikehardy The FileProvider thing is a non-breaking change and will not require any documentation change. It is just that if the library user has not defined FileProvider in their manifest, it will not cause an application crash.
As most people do not go through the manual link process and just try to do the automatic link process, which caused a lot of problems for a lot of people.

MateusAndrade:
> @mikehardy The FileProvider thing is a non-breaking change and will not require any documentation change. It is just that if the library user has not defined FileProvider in their manifest, it will not cause an application crash.
> As most people do not go through the manual link process and just try to do the automatic link process, which caused a lot of problems for a lot of people.

I think this change is valid, but as @mikehardy  have said it would be better send this on another PR? Since your PR was initially about the callback on android? Doing a lot of changes in the same PR can cause be hard to track.

example/App.js
```diff
@@ -24,6 +24,25 @@ export default class App extends Component {
     Alert.alert(`Package: ${packageSearch}`, `${isInstalled ? 'Installed' : 'Not Installed'}`);
   };
 
+  constructor(props: Props) {
```
MateusAndrade: You initialise this message on the state = {} on top of the component. Since there is already a state declaration upper

example/App.js
```diff
@@ -24,6 +24,25 @@ export default class App extends Component {
...
+    super(props);
+    this.state = {
+      result: '',
+    };
+  }
+
+  getErrorString(error: any, defaultValue?: string): string {
```
MateusAndrade: Since i updated the example, making him from the scratch i decided to remove flow to become more easier to new user understand and get up running with the library. Would be a good idea remove the flow types according with the other examples in this file?

mikehardy:
For the record, I *like* the FileProvider changes, just that it was the hold up. I had just absorbed the static class sync lock and was feeling comfortable with it, then the FileProvider stuff came in and instead of being a simple merge it went back on the review pile, behind other things :-)

kkbhav:
@MateusAndrade @mikehardy I see the problems. I know it's a lot of trouble to review multiple changes in a single PR. I will try to remove the FileProvider changes from this and will create a new PR for that.
WIll be able to complete this in a couple of hours

example/App.js
kkbhav: Removed constructor

example/App.js
kkbhav: Removed flow definition from the function

kkbhav:
@mikehardy I have resolved all the changes and removed FileProvider commits from the PR. Also, verified it as working in my internal app. Please have a look and let me know if anything else remains.

I will open a PR regarding FileProvider after this PR gets merged

index.js
```diff
@@ -60,12 +62,13 @@ class ShareSheet extends React.Component<Props> {
     return false;
   }
   render() {
+    const { style = {}, overlayStyle = {}, ...props } = this.props;
```
MateusAndrade: Why did you used that `...props`  on props?

kkbhav: Because style and overlayStyle was not meant to be passed to Overlay

mikehardy:
Thank you! I'll check them locally to make sure, but these look good to go

mikehardy:
This looked good to me - I pulled the branch and ran it in my local android emulator (and ios, but there weren't changes there) and everything seems fine.

I don't do releases in this project and it seems we haven't had one for a while so I don't know when this will go out, but I am +1 for the merge for sure. This solves a problem that was subtle and I think caused quite a few issues for people trying to use it, should be a big help to everyone

jgcmarins:
Thank you guys.
I'll try to take a huge look at react-native-share on the weekend and ship a new release as soon as possible.

mikehardy:
FWIW, it seems like everything is good and working now, just need a quick breaking changes section and major ver and goooo... :-) then wait for the issues hah

MateusAndrade:
@jgcmarins it would be a great add make a @mikehardy able to make releases on this project, since he is on almost every issue hahahahaha

jgcmarins:
done \o/

mikehardy:
Ha! I appreciate that, and I am very happy to help as I use this project, but actually making releases unless it is really systematized makes me extremely nervous. I wouldn't want to push the button until I had seen the process happen a few times :-)

mikehardy:
That said, the react-native-community's power is growing nicely and with the release of the react-native-community CircleCI orb (released yesterday IIRC, with great example use in react-native-netinfo) and a release-bot, that might really be the way to go - add a feature and if it passes CI...it ships

jgcmarins:
Awesome!
Can you share an example with us?
I can help setting up it here.

MateusAndrade:
It would be a great addition to this project. If It was possible, I would find it amazing to be able to learn a little bit about CI / CD, using this project as a basis. : D

mikehardy:
Here's the Orb itself: https://github.com/react-native-community/react-native-circleci-orb?reddit=1

I will admit to ignorance as to how to use it, as I'm a Travis CI person up till now and don't even know even CircleCI to be dangerous yet, much less helpful

But the primary force behind it - if I understand correctly - was @matt-oakes who is also the primary mover behind react-native-netinfo, which means this could probably considered a canonical usage: https://github.com/react-native-community/react-native-netinfo/blob/master/.circleci/config.yml

Then I believe the final excellent touch is to enable the semantic-release-bot hooked to publish from ci:

https://github.com/react-native-community/react-native-netinfo/blob/master/package.json#L42

https://github.com/react-native-community/react-native-netinfo/blob/master/CONTRIBUTING.md#release-process

mikehardy:
and @MateusAndrade - I submit that having done similar things (just not for react-native or CircleCI) there is nothing cooler than either your CI stopping bugs before you even knew about them or publishing software successfully after you've already left for celebratory adult beverages. Continuous Deployment is a secret superpower

kkbhav:
Great.
@jgcmarins let me know if you need anything else from my side to merge this PR

jgcmarins:
possible solution for https://github.com/react-native-community/react-native-share/issues/475

PR #489, opened 2019-04-28
chore: update iOS target to 9.0 to match react-native min version

mikehardy:
XCode complains about the deployment target in our library because it is (was) set to 7

I opened our .xcodeproj in Xcode, went to the Info tab and moved it to 9.0 to match react-native versions. The other part of the file that was touched was automatic by XCode and I have learned to not touch anything lest it explode in XCode, so I left it in

Would love if an iOS guru could verify I am not crazy, otherwise my default stance would be to +1 a merge in a few days


mikehardy:
Yes, the example :-), but please do check it yourself. It should be inconsequential but "if it is not tested, it is broken" are words to live by and we don't have ios in CI

MateusAndrade:
> Yes, the example :-), but please do check it yourself. It should be inconsequential but "if it is not tested, it is broken" are words to live by and we don't have ios in CI

Sure. I will check this on this week!

mikehardy:
Yeah - I think this is one of those thankless chores that is destined to always lag. The other modules I looked at are all over the place, with a few big ones still on 7 (react-native-fs, looking at you...). XCode doesn't complain unless it is below 8, as XCode 10.2 doesn't support below 8 anymore. We were on 7 thus generating a warning and I just figured, let's move to current react-native platform minimum 🤷‍♂

MateusAndrade:
Tested on ios Iphone X simulator and everything seems ok. Opening the project with XCode works without any problem.

@jgcmarins you can give me access to merge to master?

jgcmarins:
done \o/

MateusAndrade:
@mikehardy It looks like the base branch needs a Merge from master

mikehardy:
git co master
git pull upstream
git push
git co quiet-xcode-warnings
git rebase master
git push origin +quiet-xcode-warnings
:muscle:

Issue #484, opened 2019-04-24
new version release


MateusAndrade:
## Describe the Feature

Since the last release on npm was 6 months ago. Would be a good idea release a new version, using everything on the master? Since i started using this project, and reading some issues apparently the version on branch master is ok. But btw, it would be a good idea define a milestone before release a new version with this changes? 

## Possible Implementations

- [ ] Define a milestone, with what is needed to test before releasing a new version to npm.
- [ ] Define the changelog and breaking changes.

Since in a newbie to contributing to projects, what else would be a good idea to implement our test before releasing this version?

## Related Issues
#470

jgcmarins:
Can someone help me with release notes? Thanks.

MateusAndrade:
> Can someone help me with release notes? Thanks.

Sure! I can give a look at this today. We just need to pick the commits and describe the changes? We can comments the release notes on this issue.

jgcmarins:
Commits following [our commit lint rules](https://github.com/react-native-community/react-native-share/blob/master/commitlint.config.js) will be automatically added to release notes.
Everything else won't.
It doesn't need to be that much, just list major changes and if they have any possible breaking changes.
It's also good to list everybody involved on those changes \o/

MateusAndrade:
I run a `git log --pretty="%h - %s (%an)" v1.1.3..HEAD >> release-notes.txt` and get the following commit list:

4055990 - Implemented succeed and failed test using jest and react-native-testing-library (#408) (Vasilii)
a389cf9 - Merge pull request #495 from MateusAndrade/feature/example-md (Mateus Andrade)
9de2ff7 - Merge branch 'master' into feature/example-md (Mateus Andrade)
2ffc8ce - Adding explanation about pod install on README.MD (MateusAndrade)
f2dab69 - Adding install-pods script (MateusAndrade)
aa3a3f9 - Adding README.MD example (MateusAndrade)
47565da - Making FileProvider implementation optional (#490) (kkbhav)
9e7767c - Merge pull request #489 from mikehardy/quiet-xcode-warnings (Mateus Andrade)
f72dbe1 - chore: update iOS target to 9.0 to match react-native min version (Mike Hardy)
bb00a3f - Calling success callback at right time in Android (#482) (kkbhav)
cd231a2 - feature/updating example version (#481) (Mateus Andrade)
3849e95 - Adding react-native-community lint (#479) (Mateus Andrade)
[...]

PR #490, opened 2019-04-29
Making FileProvider implementation optional

kkbhav:
# Overview
Previously, it was necessary to implement `ShareApplication` and pass the file provider authority from `MainApplication` file. This caused a lot of linking problems for people new to android. For this, I have added a `fileProvider` in the library itself. This will make implementing FileProvider optional and will help new users to get started quickly. Also, if people would like to use their `authority` for sharing they can implement the `ShareApplication` and get it running quickly without any other implications.

# Test Plan
We can test this out by using automatic installation and base64 images to share. This will use internal provider and then we can implement FileProvider in the main application and test it out. After implementing, the library will use a user-defined provider in place of ours. And in case the user-defined provider does not implement that path, the code will try it with our own provider.
I have tested this in an internal application and it works perfectly.


android/src/main/java/cl/json/RNShareFileProvider.java
```diff
@@ -0,0 +1,7 @@
+package cl.json;
+
+import android.support.v4.content.FileProvider;
+
+public class RNShareFileProvider extends FileProvider {
```
MateusAndrade: This is correct? This class extends FileProvider but dont seems to be used on another place.

kkbhav: It is used in the manifest. Android says we cannot declare multiple providers with the same name. So we had to implement a class which extends FileProvider and then declare provider in the manifest using this class

mikehardy: That is true and what makes solving FileProvider configuration so difficult, I hadn't thought to simply extend to work around it, makes sense now that I think about it.

mikehardy:
Sorry I haven't looked at this yet @kkbhav - I like this idea in theory - just need to look it over (or someone else can) - going to be very very busy until early next week unfortunately, but I do scan my PR list periodically so it won't be lost forever even if no one else gets to it

kkbhav:
No problem.

jgcmarins:
It looks good to me.
But I've been very busy too, co

kkbhav:
Any updates?

mikehardy:
Just got back from traveling all first part of week, today is PR review day, there is a nice stack of contributions in various projects - a good problem to have :muscle:  and will try to get everything approved and merged in where I can merge. And as mentioned before I love the idea here and appreciate the effort - thanks for your patience

jgcmarins:
@kkbhav I always trust on your fixes and improvements \o/
If this one has no breaking changes, I think we are good to go

mikehardy:
Oh - I will add I did not *run* it locally. Just close review of the code. Didn't have time to pull the branch etc sorry

MateusAndrade:
> Oh - I will add I did not _run_ it locally. Just close review of the code. Didn't have time to pull the branch etc sorry

I can a try at it locally. If everything runs without problem we can proceed with the pr =D

kkbhav:
@jgcmarins Thanks. It is always good to keep improving things. There are no breaking changes, this is just making `FileProvider` implementation optional. As @mikehardy said, this will help new Android users to get started quickly and advanced users to get all advanced features out of this library.

@mikehardy I just hope I have not irritated you 😅 . Thanks for doing the hard work and helping the community. I am pushing for this so that I can start using the npm version in place of a custom one.

@MateusAndrade Thanks man for giving it a run.

kkbhav:
Please have a look at Readme.md. I have moved `FileProvider` details at the bottom of readme in the troubleshooting section as it became optional and only advanced users are gonna use it majorly.

mikehardy:
> @mikehardy I just hope I have not irritated you 😅 .

Not a bit! thanks for the work

jgcmarins:
Release?

MateusAndrade:
> Release?

SURE! I pinned a issue #484 about the release. What do you think, on discussing the release process on this issue?

mikehardy:
Yes! Even with problems, releasing is better than not releasing, for me philosophically. I don't have time to discuss now though, sorry :-(. Would have time in the next couple days to help with some urgent fix if needed though...

PR #499, opened 2019-05-20
fix android sms share default app issue

Seachaos:
# Overview
In my Android phone, the sms share is not functional.
So I use `android.provider.Telephony.Sms.getDefaultSmsPackage` to get system provider package name.

# Test Plan
when android use `Share.shareSingle` with sms, will correctly work.  ( call to internal sms app ).

### DIFF
--- a/android/src/main/java/cl/json/social/SMSShare.java
+++ b/android/src/main/java/cl/json/social/SMSShare.java
@@ -5,6 +5,7 @@ import android.content.Intent;
 import java.io.File;
 import android.os.Environment;
 import android.net.Uri;
+import android.provider.Telephony;
 
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReadableMap;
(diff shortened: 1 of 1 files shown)


MateusAndrade:
LGTM!

MateusAndrade:
@Seachaos you can try verifying your commit please? 

If you have any doubt about this, please give a look at: https://help.github.com/en/articles/checking-your-commit-and-tag-signature-verification-status

Seachaos:
> @Seachaos you can try verifying your commit please?
> 
> If you have any doubt about this, please give a look at: https://help.github.com/en/articles/checking-your-commit-and-tag-signature-verification-status

OK, I try use PGP and make commit `verified signature.`.
Is this work ?

Thanks.

mikehardy:
Interesting - if I understand this correctly, did you have a non-default SMS provider installed on your phone? If you do, and this only works with your fix (which LGTM as well), this would show how rare it is for people to change their default SMS provider or try to share by it! It would have never worked above API19 without this

Issue #504, opened 2019-05-26
v1.2.0 - link failure in iOS release mode


rollsroyc3:
## Ask your Question
I have been using this when testing on Android and iOS but now for some reason when compiling for production I keep getting the following error:

Undefined symbols for architecture x86_64:
  "_OBJC_CLASS_$_InstagramStories", referenced from:
      objc-class-ref in libRNShare.a(RNShare.o)
ld: symbol(s) not found for architecture x86_64

Any idea what could be causing this? Ive tried older commits, forking and modifying, etc.. but no luck..

MateusAndrade:
Which version of share and react-native are you using?  

You Released a new version a few hours ago. Did you tried this version?

MateusAndrade:
Also you linked the project on ios?

mikehardy:
That's strange - I just attempted to integrate the 1.2.0 version and integrate it into my day-job app and I get the same error but for armv7 architecture. I have on conclusive information but I'm not sure react-native-share is in the clear yet, I have not successfully built the release version of my app with it yet. I'll report back if/when I have more but wanted to put this out there

mikehardy:
Okay, I just moved up a layer of abstraction from fastlane (my normal release build infra) to XCode direct and I get the exact same error described in the issue. I will look for a fix really quickly but as this is a build failure in a release I want to publish what I find as I find it, even it maybe generates irritating notification traffic (sorry if so).

mikehardy:
I have taken the liberty of updating the issue title to reflect what I **think** is the most precise description of the problem. Anyone feel free to revert it if it proves to be wrong later.

mikehardy:
Okay, I just reverted to react-native-share 1.1.3 with no other changes and iOS release mode builds and archives fine. 

@MateusAndrade  @jgcmarins  it is possible to `npm unpublish` and the 1.2.0 artifact should be unpublished, it is non-functional in iOS release builds. 

I will continue looking for solutions and hopefully we will have a 1.2.1 shortly, but 1.2.0 should not be consumed.

Anyone that currently has 1.2.0 should pin 1.1.3 in their dependencies until 1.2.0 is unpublished and/or this is sorted out.

Apologies to anyone affected.

mikehardy:
[...]

PR #449, opened 2019-02-13
Instagram stories

loga4:
Hi,

It's an iOS Instagram Stories Share implementation. 

https://developers.facebook.com/docs/instagram/sharing-to-stories

PR #505, opened 2019-05-27
Revert "Instagram stories (#449)" / Fix iOS release build in 1.2.0

mikehardy:
This reverts commit e4d8ebdc1cdcf8bfec0316b205c87fade7e9f232.

As mentioned in #504 there is a link failure in release mode for iOS, and it appears to be caused by #449 

This reverts #449 until it can be repaired and iOS release builds work while it is integrated

Worth noting that #449 looks fine - it is a good feature and I am not sure what went wrong in the XCode project change - but if fails to archive in release mode at the moment

Also thankfully even though we haven't configured CI yet #290 we have astute community members that report things clearly and quickly


mikehardy:
In it's original form my revert was incomplete but I just hopped off a plane and had time to test it more fully and now it appears complete

mikehardy:
Curious what the status is on this? There's an independent report that instagram stories is the likely cause, as reverting to the commit prior archives correctly. I still see 1.2.0 on npmjs.com and this PR and the related 1.2.0 link failure are both open. So it is my current understanding that our most recent release is still unusable in production, right? Do we have a plan to merge this and do a 1.2.1 that archives? Or did I miss a different PR fix go in?

jgcmarins:
@mikehardy we haven't found the solution yet.
If reverting Instagram support solves 1.2.0, we can merge it and release 1.2.1.

mikehardy:
The PR reverting Instagram stories fixes ios archive error on  1.2.0, yes

MateusAndrade:
Since we cant find whats happening on IOS production builds using 1.2.0. i think publishing a version 1.2.1 removing the Instagram Stories is a good begining.

Then, fixing the Instagram Stories we can send a 1.2.2. We just need to adjust the docs about this change.

mikehardy:
Confirming 1.2.1 archives in release mode correctly in testing. whew!

jgcmarins:
This is awesome!
Thank you @mikehardy \o/

PR #508, opened 2019-05-27
Bugfix/ios release error

MateusAndrade:
# Overview
Fixing IOS release error after upgrading rn-share to 0.59. 

This fix: #504 .

# Test Plan
Clone the branch, and run a Production build on XCode.


jgcmarins:
`JavaScriptCore.framework` should be added on user land

Issue #507, opened 2019-05-27
Getting "No default export found in module  import/default" running eslint


jacargentina:
I'm getting this when trying to lint a file in my RN project.

![eslint](https://user-images.githubusercontent.com/517228/58434469-31c3b200-8092-11e9-9754-69ce9fb0aec5.png)

Current exports for the library:

![current](https://user-images.githubusercontent.com/517228/58434578-c75f4180-8092-11e9-934e-29767bba64a7.png)

should be converted to proper ES6 syntax https://developer.mozilla.org/es/docs/Web/JavaScript/Referencia/Sentencias/export

jgcmarins:
Maybe just refactoring from this: 
```tsx
module.exports = RNShare;
module.exports.Overlay = Overlay;
module.exports.Sheet = Sheet;
module.exports.Button = Button;
module.exports.ShareSheet = ShareSheet;
```

Into this:
```tsx
export { Overlay, Sheet, Button, ShareSheet };
export default RNShare;
```

jgcmarins:
https://github.com/react-native-community/react-native-share/pull/509

MateusAndrade:
> #509

I will give a look at this PR/ISSUE on this issue. Btw, do you think changing module.exports to export default would be a better idea?

jgcmarins:
yes, looks like it's a best practice

MateusAndrade:
@jacargentina just merged #509

PR #509, opened 2019-05-27
Chore/apply es6 exports

MateusAndrade:
# Overview
As pointed on #507 eslint throws a error because of the old export, that was using ```module.exports```. This pr fix this, applying a es6 export default and to Share and a named export to other functions

# Test Plan
Clone this repo, and run a lint and import Share and Overlay to ensure that the components are being exported correctly.


mikehardy:
I'm really not sure about this one. I've been hearing that default exports are bad, but I'm not language lawyer - I don't have the taste/wisdom/etc to say one way or another. Just the memory of recent discussion indicating default exports should be avoided.

To get rid of the default export is a breaking change though I believe, so that's not great if it is breaking.

Maybe then if getting rid of default is breaking, and this PR is not breaking but is closer to "clean" as a use of the language, then fine. I wish I were better at the module systems (there are like 3 in javascript, right?) to say one way or another

jgcmarins:
Default exports can be misleading. Named exports are safer.
This PR is certainly a breaking. But we have to fix this anyway in the future.
We can provide a code mod script, but I don't think that will be effective in this case.

MateusAndrade:
Giving a look at ```react-native-camera``` the export applied was using ES6 exports for sometime. 
 
https://github.com/react-native-community/react-native-camera/blob/18c00efd75d46ceb59615387e6c15d7469acca3b/src/index.js#L5&L6

Also ```react-native-device-info``` was using this too. As for ```export default RNShare```, if this is not applied importing lib using ```import Share from 'react-native-share'``` would give an import error.

@mikehardy, @jgcmarins  what do you guys think about adding a ```export { RNShare }```?

mikehardy:
@MateusAndrade I wish my typing skills were stronger - especially given how much I do love typescript in general - but I am not good enough to have strong opinions. That said, as a maintainer that merged someone else's PR that brought react-native-device-info to it's current status, I have heard no complaints from the userbase on the change or the current state. As feedback maybe that is useful?

MateusAndrade:
> @MateusAndrade I wish my typing skills were stronger - especially given how much I do love typescript in general - but I am not good enough to have strong opinions. That said, as a maintainer that merged someone else's PR that brought react-native-device-info to it's current status, I have heard no complaints from the userbase on the change or the current state. As feedback maybe that is useful?

Sure! I also prefer named exports, since they look more clear than using export default. Btw, i think that setting all exports to named exports will need a update on the docs. Do you think that this is a "breaking change"? :sweat_smile:

mikehardy:
I wish typing changes were in general treated as breaking changes but I feel I might be in the minority there, or at least it is not a universal feeling. I personally see no problem with large major version numbers (who cares really?) so I think breaking-change / major-version away, personally, but that's just opinion

MateusAndrade:
@jgcmarins do you think that we are good to go on this?

PR #516, opened 2019-06-04
Bugfix/fixing example

MateusAndrade:
# Overview
With the current version on the example, running a `yarn start ` if throws a error:

```
Loading dependency graph...(node:8823) UnhandledPromiseRejectionWarning: Error: jest-haste-map: Haste module naming collision:
  Duplicate module name: react-animated
  Paths: /Users/user/Development/Github/react-native-share/example/node_modules/react-native-share/node_modules/react-native/Libraries/Animated/release/package.json collides with /Users/user/Development/Github/react-native-share/example/node_modules/react-native/Libraries/Animated/release/package.json

This error is caused by `hasteImpl` returning the same name for different files.
    at setModule (/Users/user/Development/Github/react-native-share/example/node_modules/metro/node_modules/jest-haste-map/build/index.js:569:17)
    at workerReply (/Users/user/Development/Github/react-native-share/example/node_modules/metro/node_modules/jest-haste-map/build/index.js:641:9)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:8823) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:8823) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:8823) UnhandledPromiseRejectionWarning: Error: jest-haste-map: Haste module naming collision:
  Duplicate module name: react-animated
  Paths: /Users/user/Development/Github/react-native-share/example/node_modules/react-native-share/node_modules/react-native/Libraries/Animated/release/package.json collides with /Users/user/Development/Github/react-native-share/example/node_modules/react-native/Libraries/Animated/release/package.json

This error is caused by `hasteImpl` returning the same name for different files.
    at setModule (/Users/user/Development/Github/react-native-share/example/node_modules/metro/node_modules/jest-haste-map/build/index.js:569:17)
    at workerReply (/Users/user/Development/Github/react-native-share/example/node_modules/metro/node_modules/jest-haste-map/build/index.js:641:9)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:8823) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 3)
```

Adding the share dependency from git/npm fix this problem

# Test Plan
Run a `yarn && yarn start` and `react-native run-android/ios` to check if the project is running properly.


mikehardy:
:thinking: I don't like this direction, to me that makes the example out of sync. I may be wrong, and it is subtle so let me explain my thinking.

1- only way to get the example is to git clone
2- at this point you have the react-native-share code, and the example code
3- now if you npm i in the example, you'll get....potentially different code
4- examine the example for some new feature you are excited about that isn't released yet (we have released slowly, historically) and get confused...

So I like the example pointing to github master, and it seems like there must be some real problem going on

But maybe pointing to master is a terrible idea, I dunno? My taste/preference might be off

mikehardy:
That sounds so negative though, I will say that in general of course keeping the example working and everything is awesome :-). And if this does it temporarily or something, that's also okay. Better a working example than some philosophy blah blah :-)

MateusAndrade:
Im working on a way to use the local files on the project, always reflecting the master/local changes.

But, since the example os broken with the last version, i think patching it to works with the master is a better option. :)

MateusAndrade:
The problem is that using `file:../../` causes a error running the bundler with `yarn start`

mikehardy:
Yep, that's why I just use a "dev-sync" run script normally, that does a cp -R of whatever would be the package (ios, android, src, whatever) into the right spot in node modules. It's pretty rough but it always works and doesn't depend on tooling or any magic. It's still the only thing I've found that just works

MateusAndrade:
You can provide a sample code of this? Maybe It Will be better than trying some magic solution. Hahahahaha

mikehardy:
https://github.com/react-native-community/react-native-device-info/blob/master/package.json#L13

MateusAndrade:
just added a script based on react-native-device-info to sync the changes between the repo and the example

mikehardy:
> just added a script based on react-native-device-info to sync the changes between the repo and the example

The mostly braindead "...but it works" way ;-). works for me...

PR #523, opened 2019-06-13
chore: update outdated dependencies, fix prettier/flow in process

mikehardy:
# Overview

GitHub notified us that we have some security vulnerabilities

So I went and updated all of our dependencies in main repo and example

In the process I noticed that we weren't actually validating flow, so I tried it and there was a flow error related to inconsistent typing. I fixed that

Also, we weren't actually running prettier on all the files so I did that and moved the prettier config to it's own config file

Then I hooked up all the eslint/prettier/flow stuff in package.json as "validate" script so it's easier to reference in lint-staged and later via CI

The example react-native upgrade touched a few files but wasn't too serious as far as react-native upgrades go.

I'm not a flow expert, so I'm not sure if fixing the flow error is a breaking change or not. Someone using flow and importing ShareSheet as a React.Component would have to confirm

# Test Plan

`yarn validate` passes (it does now)


mikehardy:
I just noticed CI was only running "yarn lint", so I switched it to run "validate" which runs all the analyses

mikehardy:
#439 was the last time this flow type was touched perhaps @gaykov can examine the change I made to the type? I would love an opinion

MateusAndrade:
LGTM. Btw, we need to make some changes on the CI to a build and a release process on Android/Ios like that used on react-native-camera and other projects.

mikehardy:
💯  on CI changes, I'm only sad I haven't had time to try it. The react-native-community has some really good examples, netinfo is also a very good example. There is a community CircleCI orb that includes most everything and can get us to an e2e test for iOS at least and maybe android (people have been having flakiness issues with android)

I've recently discussed this with Mike Diarmid (react-native-firebase and some other) and @matt-oakes (netinfo) and the general consensus is that what netinfo is doing is the "state of the art" / current best effort for react-native-community CI, so using it as a starting point isn't a bad thing

We're also hoping to have a "template" project up soon with an example of CI integrated

This just handles static analysis a little better, and at least that's something...

MateusAndrade:
Recently i added some changes on react-native-camera ci. I was giving a look at [react-native-circleci-orb](https://github.com/react-native-community/react-native-circleci-orb) and i found this is a really good start  to discuss CI/CD on the react-native-community.

Btw, i was talking with some people on the community about creating some issues to discuss this on [.github](https://github.com/react-native-community/.github). I think I wll try to create this topics this weekend. 😄

mikehardy:
https://github.com/react-native-community/template is where it's at I think, and in template discussion - the idea being to demonstrate it in action

MateusAndrade:
Thanks!! Btw, i give a look at your PR and everything looks fine. Except by the react-native version set at the example.

mikehardy:
@MateusAndrade   `"react-native": "0.59.999999999",`  <-- what in the heck is that??

I have no idea how that happened - but I just fixed it up, sorry :-)

Issue #533, opened 2019-06-25
Share multiple images with Share.shareSingle


ondrej-merkun:
Hi,

I need to share multiple images within one app, with one dialog.
I know this is now possible with `Share.open`, but **I need to do this with just one specific social app**, so I am looking for a way to share multiple image files with one `Share.shareSingle` dialog.
How could this be done?

Thanks in advance.

jgcmarins:
Yes, it is possible. It's on developer land.
You can add an if statement and check it before call the share function.

merkur0:
> You can add an if statement and check it before call the share function.

What do you mean by that? How would an if statement help me with sharing multiple image files using a method that seems to only support one image file?

jgcmarins:
Check the example:
https://github.com/react-native-community/react-native-share/blob/05c2b6a1aeb74853ef16265f690b3ba48cd0198f/example/App.js#L58-L78

merkur0:
> Check the example:
> 
> https://github.com/react-native-community/react-native-share/blob/05c2b6a1aeb74853ef16265f690b3ba48cd0198f/example/App.js#L58-L78

I am aware of this, however, as I mentioned in my original post, I need to share the image files within one app (just like with Share.shareSingle). Share.open can't do that, therefore I can't use this example.

ronal2do:
@merkur0 I think you can if you look to types 
```ts
declare namespace Share {
    function shareSingle(
        options: Options & { social: SupportedSocialApps }
    ): Promise<ShareSingleReturn>;
}

interface Options {
    url?: string;
    urls?: string[];
    type?: string;
    message?: string;
    title?: string;
    subject?: string;
    excludedActivityTypes?: string;
    failOnCancel?: boolean;
    showAppsToView?: boolean;
}

type SupportedSocialApps =
    | "facebook"
    | "pagesmanager"
    | "twitter"
    | "whatsapp"
    | "instagram"
    | "googleplus"
    | "email";
```

kkbhav:
Multiple image share will not work with shareSingle on iOS. Somebody will have to raise PR for this feature. For Android, it is already implemented.

xliank:
Here is my script, 
Simple explain, I need to download the pictures and convert to base64 using rn-fetch-blob.
I theory you can share files/pictures/etc with some library.

> import RNFetchBlob from "rn-fetch-blob";
import Share from "react-native-share";
import {
  PermissionsAndroid,
  ToastAndroid,
  Alert,
[...]

Issue #389, opened 2018-10-24
iOS: shareSingle on instagram open camera even if photo url (base64) is passed to url parameter


atlkuk:
### Steps to reproduce
```
RNFetchBlob.fs.readFile(path, 'base64')
.then(result => {
    const shareOptions = {
         url: `data:image/jpeg;base64,${result}`,
         social: Share.Social.INSTAGRAM
    };
Share.shareSingle(shareOptions)
.then((data) => console.log(data))
.catch((error) => console.log(error));
```

### Expected behaviour
Instagram should show sharing with my photo already loaded

### Actual behaviour
Instagram show camera to catch a new photo

### Environment
- **React Native version**: 0.57.2
- **React Native platform + platform version**: iOS 12.0.1

### react-native-share
**Version**: 2.4.2

yeomann:
maybe it is something Instagram default behavior now 🤔 ?

atlkuk:
I don't think so. On android I can share my photo as before.

ahilles107:
i have this same issue - no solution yet ;/

Desintegrator:
same

ethanyuwang:
same issue here

ethanyuwang:
Has anyone solved this?

Vikas-Sharma7:
Still happening.

MateusAndrade:
Sharingan with a URL works?

pragadeeshk:
Same issue still happening

> Sharing with a URL works?

It doesn't even open the share window.

jgcmarins:
is this related: https://github.com/react-native-community/react-native-share/issues/232
?

MateusAndrade:
> is #232 related?

Its possible, when i was working with rn-share trying to send a base64 was problematic with some apps. I decided instead, to send the relative image Path. By doing this, It Works on most of the cases.

Issue #388, opened 2018-10-24
Although Instagram App installed my device, share is not working. It is opening to App Store.


buraksak35:
### Expected behaviour
Open directly instagram app

### Actual behaviour
Open app store

### Environment
  React Native Environment Info:
    System:
      OS: macOS 10.14
    Binaries:
      Node: 10.11.0 - /usr/local/bin/node
      Yarn: 1.10.1 - /usr/local/bin/yarn
      npm: 6.4.1 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.0, macOS 10.14, tvOS 12.0, watchOS 5.0
      Android SDK:
        Build Tools: 23.0.1, 26.0.1, 26.0.2, 26.0.3, 27.0.0, 27.0.3, 28.0.2
        API Levels: 22, 23, 26, 27, 28
    npmPackages:
      react: 16.4.1 => 16.4.1
      react-native: 0.56.0 => 0.56.0
    npmGlobalPackages:
      react-native-cli: 2.0.1
      react-native-git-upgrade: 0.2.7

### react-native-share
"react-native-share": "^1.1.2",

![ekran resmi 2018-10-24 11 34 07](https://user-images.githubusercontent.com/39245976/47417528-d4c0f000-d780-11e8-93b1-41100152bcda.png)

mikehardy:
Which was installed first? The app using RNS or instagram? I assume from the code snippet it is iOS? What device or emulator version and iOS version?

buraksak35:
> Which was installed first? The app using RNS or instagram? I assume from the code snippet it is iOS? What device or emulator version and iOS version?

İphone 5s , IOS 11.I have a button as "Share to instagram" when I click instagram should open . But appstore is opening

mgiorgi-github:
Same issue, all app (Facebook, Instagram and whatsapp) are not opened but the share module open app to Apple store.
I’ve tryed the code example and use “Share ui component” link

mgiorgi-github:
i've to pass image with local url path (file://) or base64?
Instagram open last image of my gallery, not the image i've selected, Facebook give error message (wrong url)

jgcmarins:
it works for both approaches
which error message are you experiencing when sharing at Facebook?

mgiorgi-github:
With my iPhone 6 iOS 12.1:
Instagram now open app but with wrong image
[...]

Issue #232, opened 2018-03-28
Share single iOS base64 img problem


melihberberolu:
Share single method works on android perfectly.

iOS sharing whatsapp, fb, twitter does not show my base64 img

`const storedImg = `data:image/png;base64,${resp.base64Img}`;
  Share.shareSingle({
    title: _('Tone Tone High Score Share'),
    message: "https://apple.co/2pmdyte",
    url: storedImg,
    subject: "https://apple.co/2pmdyte",
    social
  }).then(res => res)
    .catch(e => e);`

jgcmarins:
@melihberberolu could you provide more details about your problem?

melihberberolu:
I'm trying to share image on social media.(Facebook, whatsapp, twitter)
My sharing is works perfectly in android platform.
But my message and base64 string not transform to image.
It's show url + base64 string.

jgcmarins:
@melihberberolu could you provide some code example? Thanks.

melihberberolu:
Yes sure @jgcmarins. You can share any example with me for sharing base64Img on iOS.
```
const share = (base64Img, social = 'facebook') => {
  Share.shareSingle({
    title: "Tone Tone High Score Share",
    message: "https://apple.co/2pmdyte",
    url: `data:image/png;base64,${base64Img}`,
    subject: "https://apple.co/2pmdyte",
    social
 }).then(res => res)
    .catch(e => e);
}
```

phillmill:
@melihberberolu did you solve?

ruimfernandes:
@melihberberolu are you sure your image is well converted into base64? And are you sure your image is **png** or it can have another extension?

Try to paste the result of `data:image/png;base64,${base64Img}` at the website and check if the image is rendered.
https://codebeautify.org/base64-to-image-converter

I'm using `Share.open` and the only problem I have while using base64 it's with Whatsapp
```
        base64Data = `data:image/${type};base64,` + base64Data;
        Share.open({
          url: base64Data
        });
```

never00miss:
did anyone solve this?

jgcmarins:
We have a troubleshooting section about sharing base64 files: https://github.com/react-native-community/react-native-share#troubleshooting

atdparvaiz:
The issue still persists, did anybody solved it?

jgcmarins:
We did a new release, can you try it? v1.2.0.
Thank you.

vivekvijayakrishnan:
[...]

PR #535, opened 2019-07-01
Facebook, Twitter and Instagram shareSingle - Fix on iOS 11

IbrahimSulai:
* Fix for Facebook and Twitter shareSingle is not working on iOS 11.
* Handled the image share option on Instagram.

### DIFF
--- a/ios/InstagramShare.m
+++ b/ios/InstagramShare.m
@@ -7,6 +7,7 @@
 
 #import "InstagramShare.h"
 #import <AVFoundation/AVFoundation.h>
+@import Photos;
 
 @implementation InstagramShare
     RCT_EXPORT_MODULE();
--- a/ios/InstagramShare.m
+++ b/ios/InstagramShare.m
@@ -50,4 +50,74 @@
     } 
 }
 
+- (void)shareSingleImage:(NSDictionary *)options
+         failureCallback:(RCTResponseErrorBlock)failureCallback
+         successCallback:(RCTResponseSenderBlock)successCallback {
+    
+    UIImage *image;
+    NSURL *imageURL = [RCTConvert NSURL:options[@"url"]];
+    if (imageURL) {
+        if (imageURL.fileURL || [imageURL.scheme.lowercaseString isEqualToString:@"data"]) {
+            NSError *error;
+            NSData *data = [NSData dataWithContentsOfURL:imageURL
+                                                 options:(NSDataReadingOptions)0
+                                                   error:&error];
+            if (!data) {
+                failureCallback(error);
+                return;
+            }
+            image = [UIImage imageWithData: data];
+            [self savePictureAndOpenInstagram: image
+                              failureCallback: failureCallback
+                              successCallback: successCallback];
+        }
+    } else {
+        [[UIApplication sharedApplication] openURL: [NSURL URLWithString:@"instagram://camera"]];
+        successCallback(@[]);
+    }
+}
+
+-(void)savePictureAndOpenInstagram:(UIImage *)base64Image
+                   failureCallback:(RCTResponseErrorBlock)failureCallback
+                   successCallback:(RCTResponseSenderBlock)successCallback {
+    
+    NSURL *URL = [self fileURLWithTemporaryImageData:UIImageJPEGRepresentation(base64Image, 0.9)];
+    __block PHAssetChangeRequest *_mChangeRequest = nil;
+    __block PHObjectPlaceholder *placeholder;
+    
+    [[PHPhotoLibrary sharedPhotoLibrary] performChanges:^{
+        
+        NSData *pngData = [NSData dataWithContentsOfURL:URL];
+        UIImage *image = [UIImage imageWithData:pngData];
+        _mChangeRequest = [PHAssetChangeRequest creationRequestForAssetFromImage:image];
+        placeholder = _mChangeRequest.placeholderForCreatedAsset;
+    } completionHandler:^(BOOL success, NSError *error) {
+        
+        if (success) {
+            NSURL *instagramURL = [NSURL URLWithString:[NSString stringWithFormat:@"instagram://library?LocalIdentifier=\%@", [placeholder localIdentifier]]];
+            
+            if ([[UIApplication sharedApplication] canOpenURL:instagramURL]) {
+                [[UIApplication sharedApplication] openURL:instagramURL options:@{} completionHandler:NULL];
+                if (successCallback != NULL) {
+                    successCallback(@[]);
+                }
+            }
+        }
+        else {
+            //Error while writing
+            if (failureCallback != NULL) {
+                failureCallback(error);
+            }
+        }
+    }];
+}
+
+- (NSURL *)fileURLWithTemporaryImageData:(NSData *)data {
+    NSString *writePath = [NSTemporaryDirectory() stringByAppendingPathComponent:@"instagram.ig"];
+    if (![data writeToFile:writePath atomically:YES]) {
+        return nil;
+    }
+    return [NSURL fileURLWithPath:writePath];
+}
+
 @end
--- a/ios/GenericShare.h
+++ b/ios/GenericShare.h
@@ -52,4 +52,5 @@
 @interface GenericShare : NSObject <RCTBridgeModule>
 
 - (void *) shareSingle:(NSDictionary *)options failureCallback:(RCTResponseErrorBlock)failureCallback successCallback:(RCTResponseSenderBlock)successCallback serviceType:(NSString*)serviceType;
+    inAppBaseUrl:(NSString *)inAppBaseUrl;
 @end
--- a/ios/GenericShare.h
+++ b/ios/GenericShare.h
@@ -51,6 +51,6 @@
 #endif
 @interface GenericShare : NSObject <RCTBridgeModule>
 
-- (void *) shareSingle:(NSDictionary *)options failureCallback:(RCTResponseErrorBlock)failureCallback successCallback:(RCTResponseSenderBlock)successCallback serviceType:(NSString*)serviceType;
+- (void *) shareSingle:(NSDictionary *)options failureCallback:(RCTResponseErrorBlock)failureCallback successCallback:(RCTResponseSenderBlock)successCallback serviceType:(NSString*)serviceType
     inAppBaseUrl:(NSString *)inAppBaseUrl;
 @end
--- a/ios/GenericShare.m
+++ b/ios/GenericShare.m
@@ -13,10 +13,11 @@
 - (void)shareSingle:(NSDictionary *)options
     failureCallback:(RCTResponseErrorBlock)failureCallback
     successCallback:(RCTResponseSenderBlock)successCallback
-    serviceType:(NSString*)serviceType {
+    serviceType:(NSString*)serviceType
+    inAppBaseUrl:(NSString *)inAppBaseUrl {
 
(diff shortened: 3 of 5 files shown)


IbrahimSulai:
The above fix will solve the below issues:

https://github.com/react-native-community/react-native-share/issues/232

https://github.com/react-native-community/react-native-share/issues/174

https://github.com/react-native-community/react-native-share/issues/389

https://github.com/react-native-community/react-native-share/issues/388

MateusAndrade:
WOW! This looks really good! Im running out of time this week to review this. But doing a overview its LGTM! 👍 

@mikehardy you can give a look at this?

mikehardy:
I love the idea of fixing a pile of bugs for people, and this looks like good code - I read through it - the problem (for me, it’s my problem) is that I’m not an iOS developer, I can only pretend I am because of react-native. I’m your Android / Java review guy :-). I wish I could look at this and say “yes!” but I just don’t have the skills.

What I do have is some basic quality control questions though - @IbrahimSulai can you help me out?

1. react-native supports back to ios9 through current (12.x) currently, and our module is in harmony with that. Have you tested these changes on ios9, 10 and 12? I assume 11 yes since it’s in the PR name…
1. Are you using this in production (or is it in testing for release) in an app? How many users does the app have, and does it spread across those ios levels?
1. Can you handle any follow-on issues with this specific PR if it causes problems? I frequently propose a PR in a repo and it’s 99% and has one crash bug for instance, and I’ll chase my first PR with a fix, then all is quiet. Will you listen for problems for a bit to make sure it lands well?

I ask those things because even in situations where I don’t personally have the skills, if all those answers are a “yes”, I am biased towards giving it a shot. Sometimes things break, but we all move forward, and it’s awesome to get PRs with fixes :-)

IbrahimSulai:
> I love the idea of fixing a pile of bugs for people, and this looks like good code - I read through it - the problem (for me, it’s my problem) is that I’m not an iOS developer, I can only pretend I am because of react-native. I’m your Android / Java review guy :-). I wish I could look at this and say “yes!” but I just don’t have the skills.

> 
> What I do have is some basic quality control questions though - @IbrahimSulai can you help me out?
> 
> 1. react-native supports back to ios9 through current (12.x) currently, and our module is in harmony with that. Have you tested these changes on ios9, 10 and 12? I assume 11 yes since it’s in the PR name…

> [@IbrahimSulai ] **I have tested in iOS 11 and iOS 12  since 96% of iOS devices are running in iOS 11 and above. Also, I don't have a lower version device with me for testing as of now. Expecting these changes won't break anything in lower versions and ready to take up any issues related to lower versions.**

> 2. Are you using this in production (or is it in testing for release) in an app? How many users does the app have, and does it spread across those ios levels?

> [@IbrahimSulai ] **Yes, our application is in production and the changes are in the beta release. The minimum version of the app supports is iOS 11.**

> 3. Can you handle any follow-on issues with this specific PR if it causes problems? I frequently propose a PR in a repo and it’s 99% and has one crash bug for instance, and I’ll chase my first PR with a fix, then all is quiet. Will you listen for problems for a bit to make sure it lands well?

>[@IbrahimSulai ] **Sure @mikehardy , I will be continuously available and support the RN community especially this share-component.**

> I ask those things because even in situations where I don’t personally have the skills, if all those answers are a “yes”, I am biased towards giving it a shot. Sometimes things break, but we all move forward, and it’s awesome to get PRs with fixes :-)

IbrahimSulai:
@mikehardy - I have answered the questions above in inline. Please review and merge the same. Thanks

mikehardy:
I’m not satisifed by the “have not tested below 11” and “our app doesn’t use 11” answers - these items are subtle, it needs testing on 9 and 10 to make sure, you don’t need lower level devices I believe this stuff is simulator capable and you can run iOS 9 and iOS 10 simulators yes? Knowing that the code itself is in production sounds great though, that means for your intended audience (11+) you know the code works and we can have more confidence in it.

And very glad of course that you’ll listen in case something goes wrong.

If someone (yourself or perhaps someone on the linked issues) could affirm it works on ios 9 and 10 then we may be ready

mikehardy:
(one reason I stick on testing the actual APIs when testing something version specific - notifications for instance is massively different between 9 and 10 - sure, not many people use it, but a test is easier than responding to bug reports from the few that *must* have it for 9)

IbrahimSulai:
@mikehardy Yes, that makes sense to me. I don't have iOS 9 devices with me at present. I will try to get one for testing. Any help from the community for testing on iOS 9 is appreciated.

mikehardy:
@IbrahimSulai the simulator should suffice? You can still get ios 9 and 10 images for the simulator (I use them for my app)

jgcmarins:
Awesome work guys.
I don't see any problem merging this one.
Most Apple users are at iOS 11 or above: https://developer.apple.com/support/app-store/

IbrahimSulai:
@mikehardy @MateusAndrade 
I have tested the implemented code in the simulator iOS 9 and iOS 10. It's working fine as expected and not breaking the existing functionality.

For testing purpose, I have created the below repository, which includes my current implementation also. Please have a look and try.
https://github.com/IbrahimSulai/react-native-share-test/tree/master

mikehardy:
@jgcmarins my app has many ios 10 and even a 9 user. Most don't I agree but why breaking change and drop working platforms when a test will do? React native supports to 9 and consistency has its own value as well. Modern only APIs should be version checked instead of just breaking old ios until the ecosystem moves in my opinion 

@ibrahimsulai thank for checking 😁

IbrahimSulai:
@mikehardy , If there are no more things pending w.r.t this PR, can we get this merged? Thanks.

mikehardy:
Done! Great work thank you!

IbrahimSulai:
Thanks @mikehardy.

bd-arc:
Hi @IbrahimSulai,

Thanks a lot for your work on those issues!

Unless I haven't understood the proper way of sharing images, it seems that there still is a problem with base64 images.

I'm using your update and I've implemented something along the lines of what @xliank does [here](https://github.com/react-native-community/react-native-share/issues/533#issuecomment-508459727). But when I pass my base64 image as the `url` parameter in `Share.shareSingle()`: 

1. for Instagram it opens the App Store page instead of the app (whereas it opens the app if I ditch the `url` parameter)
2. for Facebook it makes the website crash (haven't tested with the app installed yet).

Do you have any idea about what could be going wrong?

MateusAndrade:
@bd-arc you tried using your file path insteatd of passing a base64? Since i started using rn-share, sharing base64 files looks problematic on some cases.

bd-arc:
@MateusAndrade Thank you for the suggestion!

I did actually, but it didn't work either. I used the temporary file path retrieved with `RNFetchBlob.config({ fileCache: true }).then((res: any) => { return res.data; });` to no avail.

Do you know of a better way to retrieve it?

I'm considering downloading the file to the camera roll and then getting the local identifier, as per @itsdante's [suggestion](https://github.com/react-native-community/react-native-share/issues/388#issuecomment-464363636). But this is an overkill that requires specific user permissions and will probably not work on Android...

IbrahimSulai:
@bd-arc - I have created a below sample for passing base64 image directly in shareSingle and its working perfectly for Facebook, Instagram and Twitter.

https://github.com/IbrahimSulai/react-native-share-test/tree/master

In your case - Please try to get the base64 Image from the url separately and give the retrieved base64 Image to the url parameter in the Share.shareSingle().

Please share your sample if you are still not able to resolve your problem. Thanks.

bd-arc:
Thanks a lot @IbrahimSulai for taking the time to put an example together! I really appreciate it 👍

I'm going to use it as a starting point and will get back to you when I've either made it work or put together a test case.

bd-arc:
First of all, thanks again for tackling this issues @IbrahimSulai. Indeed, there wasn't any problem with your code.

I've finally managed to properly share remote images by converting them to base64 images. I'll share some of my findings below since it might prove useful to others too.

---

⚠️ The following is for **iOS only**. Didn't have a chance to try it on Android yet.

---

There were two requirements to make this work that I wasn't aware of and about which I didn't read in any of the issues I browsed:

1. **Your app's `Info.plist` must contain an `NSPhotoLibraryUsageDescription` key**, otherwise it will crash as soon as you try to share your image — on a real device at least, because the simulator stands still. Whether you're downloading the image first, importing a local image64 file or reading a remote URL, you need to include this permission. This wasn't obvious to me, at all.

2. **You need to add every scheme you intend to use in an `LSApplicationQueriesSchemes` key**. This is something I completely overlooked. As an example, here's what I'm using for Instagram and Facebook:

```
<key>LSApplicationQueriesSchemes</key>
<array>
    <string>fb</string>
    <string>fbapi</string>
    <string>fbauth2</string>
    <string>fbshareextension</string>
    <string>fb-messenger-api</string>
    <string>instagram</string>
    <string>instagram-stories</string>
</array>
```

---

And here's how I handle image retrieval/conversion:

```
private share (socialRef?: 'FACEBOOK' | 'INSTAGRAM') {
    RNFetchBlob.fetch('GET', this.props.shareImageUrl)
        .then((res: any) => {
            Share.shareSingle({
                message: 'Your message',
                social: Share.Social[socialRef],
                url:  'data:image/jpeg;base64,' + res.base64(),
                type: 'image/jpeg',
                failOnCancel: false
            })
            .catch(() => {
                // Catch error
            });
        })
        .catch(() => {
            // Catch error
        });
}
```

---

Notes:

* This code is far from perfect since it will create a new image on every call. The proper way to do so would be saving the image to a specific location, checking if it already exists on subsequent calls and loading it from the file system if need be. [`react-native-fs`](https://github.com/itinance/react-native-fs) and [`rn-fetch-blob`](https://github.com/joltup/rn-fetch-blob) are great plugins to help with that.

* If the Facebook app isn't installed, the plugin will default on opening the browser. **With a base64 image, it will actually crash the Facebook website**. I recommend checking whether the app is installed or not thanks to a plugin like [this one](https://github.com/redpandatronicsuk/react-native-check-app-install).

Hope this helps!

MateusAndrade:
@bd-arc thanks for sharing your solution. Would you mind to send a PR updating the README.MD showing this? I see that this is a common issue to people that use react-native-share.

Thanks again for sharing your solution. :smiley:

IbrahimSulai:
@bd-arc @MateusAndrade  - Great to hear that my PR is working as expected. 
Thanks for providing steps to converting the remote image into a base64 image. 
I am always open to take any issues coming in future regarding this functionality. I hope it should not be any :)
Thanks

bd-arc:
Here you are @MateusAndrade: #552 ;-)

ios/InstagramShare.m
```diff
@@ -50,4 +51,74 @@ - (void)shareSingle:(NSDictionary *)options
...
+        [[UIApplication sharedApplication] openURL: [NSURL URLWithString:@"instagram://camera"]];
+        successCallback(@[]);
+    }
+}
+
+-(void)savePictureAndOpenInstagram:(UIImage *)base64Image
+                   failureCallback:(RCTResponseErrorBlock)failureCallback
+                   successCallback:(RCTResponseSenderBlock)successCallback {
+    
+    NSURL *URL = [self fileURLWithTemporaryImageData:UIImageJPEGRepresentation(base64Image, 0.9)];
```
ProdigySim, line 85: Is this recompressing the image before sending it to Instagram?

jgcmarins: it seems so

Issue #549, opened 2019-07-17
Android release build doesn't work with react-native@0.60 (AndroidX support is needed)


oleksandr-dziuban:
### Steps to reproduce
1. Init new fresh project with latest react-native@0.60.3
2. Install latest react-native-share (with autolinking)
3. Run RELEASE Android build assembling

### Expected behaviour
Release Android build is assembled.

### Actual behaviour
Release Android build is failed.

### Error

```
error: package android.support.v4.content does not exist
[05:45:01]: ▸ import android.support.v4.content.FileProvider;
[05:45:01]: ▸ ^
[05:45:01]: ▸ /Users/distiller/mobile-web-app/node_modules/react-native-share/android/src/main/java/cl/json/RNShareFileProvider.java:5: error: cannot find symbol
[05:45:01]: ▸ public class RNShareFileProvider extends FileProvider {
[05:45:01]: ▸ ^
[05:45:01]: ▸ symbol: class FileProvider
[05:45:01]: ▸ /Users/distiller/mobile-web-app/node_modules/react-native-share/android/src/main/java/cl/json/RNShareModule.java:7: error: package android.support.annotation does not exist
[05:45:01]: ▸ import android.support.annotation.Nullable;
[05:45:01]: ▸ ^
[05:45:01]: ▸ /Users/distiller/mobile-web-app/node_modules/react-native-share/android/src/main/java/cl/json/RNSharePathUtil.java:12: error: package android.support.annotation does not exist
[05:45:01]: ▸ import android.support.annotation.NonNull;
[05:45:01]: ▸ ^
[05:45:01]: ▸ /Users/distiller/mobile-web-app/node_modules/react-native-share/android/src/main/java/cl/json/RNSharePathUtil.java:13: error: package android.support.v4.content does not exist
[05:45:01]: ▸ import android.support.v4.content.CursorLoader;
[05:45:01]: ▸ ^
[05:45:01]: ▸ /Users/distiller/mobile-web-app/node_modules/react-native-share/android/src/main/java/cl/json/RNSharePathUtil.java:14: error: package android.support.v4.content does not exist
[05:45:01]: ▸ import android.support.v4.content.FileProvider;
[05:45:01]: ▸ ^
[05:45:01]: ▸ /Users/distiller/mobile-web-app/node_modules/react-native-share/android/src/main/java/cl/json/RNShareModule.java:126: error: cannot find symbol
[...]

PR #550, opened 2019-07-18
AndroidX libraries and react-native@0.60.0+ support

oleksandr-dziuban:
- AndroidX libraries support added
- Fixed Android release build issues with old File Provider libraries
- react-native@0.60.0+ support and auto linking support added
- Updated ReadMe file with usage for eact-native@0.60.0+

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,9 @@ cd ios && pod install && cd ../
 ___
 
 ### If you are using `react-native` <= 0.59.10 please use `react-native-share` <= 1.2.1:
+If you are having any problems with this library, or need to use >= 2.0.0 please refer to: [jetifier](https://github.com/mikehardy/jetifier#to-reverse-jetify--convert-node_modules-dependencies-to-support-libraries).
 
+After installing jetifier, runs a ```npx jetify -r``` and test if this works by running a ```react-native run-android```.
 ## Automatic Way
 
 ---
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ cd ios && pod install && cd ../
 ```
 ___
 
-### If you are using `react-native` >= 0.60.0 please use `react-native-share` <= 1.2.1:
+### If you are using `react-native` <= 0.59.10 please use `react-native-share` <= 1.2.1:
 
 ## Automatic Way
 
--- a/README.md
+++ b/README.md
@@ -13,6 +13,30 @@ This gives you the power to prioritize our work and support the project contribu
 # Getting started
 ---
 
+### If you are using `react-native` >= 0.60.0 please use `react-native-share` >= 2.0.0
+
+## Automatic Way
+---
+``` 
+yarn add react-native-share
+```
+
+or if you're using npm
+``` 
+npm install react-native-share --save
+```
+---
+
+#### Important:
+Linking is not needed anymore. ``react-native@0.60.0+`` supports dependencies auto linking.
+For iOS you also need additional step to install auto linked Pods (Cocoapods should be installed):
+``` 
+cd ios && pod install && cd ../
+```
+___
+
+### If you are using `react-native` >= 0.60.0 please use `react-native-share` <= 1.2.1:
+
 ## Automatic Way
 
 ---
@@ -324,7 +348,7 @@ static sharePDFWithAndroid(fileUrl, type) {
     ```xml
     <application>
         <provider
-            android:name="android.support.v4.content.FileProvider"
+            android:name="androidx.core.content.FileProvider"
             android:authorities="${applicationId}.provider"
             android:grantUriPermissions="true"
             android:exported="false">
--- a/android/src/main/java/cl/json/RNShareFileProvider.java
+++ b/android/src/main/java/cl/json/RNShareFileProvider.java
@@ -1,6 +1,6 @@
 package cl.json;
 
-import android.support.v4.content.FileProvider;
+import androidx.core.content.FileProvider;
 
 public class RNShareFileProvider extends FileProvider {
 
--- a/android/src/main/java/cl/json/RNShareModule.java
+++ b/android/src/main/java/cl/json/RNShareModule.java
@@ -4,7 +4,7 @@ import android.app.Activity;
 import android.content.ActivityNotFoundException;
 import android.content.Intent;
 import android.net.Uri;
-import android.support.annotation.Nullable;
+import androidx.annotation.Nullable;
 
 import com.facebook.react.bridge.ActivityEventListener;
 import com.facebook.react.bridge.ReactApplicationContext;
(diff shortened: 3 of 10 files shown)


jgcmarins:
Awesome!
We're facing flow errors `Check for errors in the example app using Flow`. Could you take a look please?
I would also ask you to update [example](https://github.com/react-native-community/react-native-share/tree/master/example)

oleksandr-dziuban:
> example

Sure, working on it now, forgot to update example

oleksandr-dziuban:
> Awesome!
> We're facing flow errors `Check for errors in the example app using Flow`. Could you take a look please?
> 
> Flow Errors (click to expand)
> I would also ask you to update [example](https://github.com/react-native-community/react-native-share/tree/master/example)

@jgcmarins Fixed, please check

MateusAndrade:
AWESOME! :tada:  :tada: 

Thanks for helping us @oleksandr-dziuban. I gave a quick look, and It LGTM.

oleksandr-dziuban:
> AWESOME! 🎉 🎉
> 
> Thanks for helping us @oleksandr-dziuban. I gave a quick look, and It LGTM.
> 
> Im going to give a quick test at this PR as soon as possible, so we can Merge this. 😄

Thanks, waiting publishing version 2.0.0 👍

oleksandr-dziuban:
@MateusAndrade Hello, I have checked it with fresh project and even with production live project in debug and release mode. No errors, Android/iOS is assembling correctly

oleksandr-dziuban:
> Regarding [this](https://github.com/react-native-community/react-native-share/issues/549#issuecomment-513048235) comment, we should add instructions about [jetifier](https://github.com/mikehardy/jetifier) on README, as well.

Jetifier is always enabled if user migrated to react-native@0.60+. It is added as enable.jetifier=true property in gradle settings.

mikehardy:
@oleksandr-dziuban I said:

```
and then people that are stuck on RN0.59 should be advised to use jetifier -r as a package.json postinstall.
```

Your comment does not address that.

There are two jetifiers. There is the Java-level gradle tool you reference. You are correct it is always on for RN0.60, and the react-native-community/cli authors even worked with me to add the second tool - the javascript-level node_modules jetifier - to the CLI so it is on by default as well.

But what do you do when you are stuck on RN0.59 and you want to use the latest react-native-share after your breaking change?

Consider that.

That is what `jetifier -r` is for. It reverses the mapping then the transition is easy (by making modern library versions compatible with android support libraries) for people that can't move quite yet.

That's what I meant in my comment

oleksandr-dziuban:
> @oleksandr-dziuban I said:
> 
> ```
> and then people that are stuck on RN0.59 should be advised to use jetifier -r as a package.json postinstall.
> ```
> 
> Your comment does not address that.
> 
> There are two jetifiers. There is the Java-level gradle tool you reference. You are correct it is always on for RN0.60, and the react-native-community/cli authors even worked with me to add the second tool - the javascript-level node_modules jetifier - to the CLI so it is on by default as well.
> 
> But what do you do when you are stuck on RN0.59 and you want to use the latest react-native-share after your breaking change?
> 
> Consider that.
> 
> That is what `jetifier -r` is for. It reverses the mapping then the transition is easy (by making modern library versions compatible with android support libraries) for people that can't move quite yet.
> 
> That's what I meant in my comment

Ok, I see, thanks. 
So what are the next steps? You don't want to merge this PR?

mikehardy:
I have been clear on my preference - the PR is the right direction, assuming it's good (I've not tested it), merging an AndroidX transition PR is the right idea.

The next step is just to update documentation for people still on react-native 0.59, tell them they will need `jetifier -r` in their postinstall after this. This will be a common recommendation across lots of libraries

MateusAndrade:
> I have been clear on my preference - the PR is the right direction, assuming it's good (I've not tested it), merging an AndroidX transition PR is the right idea.
> 
> The next step is just to update documentation for people still on react-native 0.59, tell them they will need `jetifier -r` in their postinstall after this. This will be a common recommendation across lots of libraries

I can handle this update and the test. :smile: 

But how someone using rn 0.59 would need to updated their project? The will need to enable jetifier? This is not clear to me.

mikehardy:
If you read this you'll understand, and you can link to it, https://github.com/mikehardy/jetifier/blob/master/README.md#to-reverse-jetify--convert-node_modules-dependencies-to-support-libraries

MateusAndrade:
Just tested on a project using the following: 

```json
{
    "react-native": "0.59.2",
    "react-native-animatable": "^1.3.2",
    "react-native-camera": "^1.10.0",
    "react-native-confirmation-code-input": "^1.0.4",
    "react-native-firebase": "5.2.1",
    "react-native-gesture-handler": "^1.0.15",
    "react-native-keyboard-aware-scroll-view": "^0.8.0",
    "react-native-linear-gradient": "^2.5.3",
    "react-native-masked-text": "^1.10.1",
    "react-native-orientation-locker": "^1.1.2",
    "react-native-picker-select": "^6.0.0",
    "react-native-share": "git+https://github.com/oleksandr-dziuban/react-native-share",
    "react-native-splash-screen": "^3.2.0",
    "react-native-tab-view": "^1.3.1",
    "react-native-vector-icons": "^6.2.0",
    "react-native-view-shot": "^2.6.0",
    "react-navigation": "^3.0.9"
}
```

And it works for dev mode using ```react-native run-android``` and for release without any problem. i just needed to run a ```npx jetify -r``` after running a ```yarn``` and it build for both modes like a charm.  I also tested the share functions using both dev/release build and it works perfectly. :tada: :tada: :tada: 

Btw, thanks @mikehardy for saving our lifes with jetifier. :smile:

README.md
```diff
@@ -13,6 +13,30 @@ This gives you the power to prioritize our work and support the project contribu
...
+#### Important:
+Linking is not needed anymore. ``react-native@0.60.0+`` supports dependencies auto linking.
+For iOS you also need additional step to install auto linked Pods (Cocoapods should be installed):
+``` 
+cd ios && pod install && cd ../
+```
+___
+
+### If you are using `react-native` <= 0.59.10 please use `react-native-share` <= 1.2.1:
+
```
MateusAndrade: Would be interesting warning about the possible problems using this PR. Adding something like: 

```suggestion
If you are having any problems with this library, or need to use >= 2.0.0 please refer to: [jetifier](https://github.com/mikehardy/jetifier#to-reverse-jetify--convert-node_modules-dependencies-to-support-libraries).

After installing jetifier, runs a ```npx jetify -r``` and test if this works by running a ```react-native run-android```.
```

MateusAndrade:
Its looks ok for me now. Do you think we can merge this @mikehardy?

MateusAndrade:
Thanks for helping with that @oleksandr-dziuban! 

@mikehardy  what do you think about pin a issue wih the possible problems related to this update?

mikehardy:
@MateusAndrade maybe? :-) - in react-native-firebase we pinned one because the issue volume was crazy. I would leave it to you, if you notice you're spending a lot of time on AndroidX chatter, pin one. Can't really hurt. Nothing else super-hot in this project I think so it's not as if the pin area is too full?

PR #579, opened 2019-09-06
Update Gradle wrapper to 5.6.2

friederbluemle:
# Overview

- Gradle wrapper `5.6.2`

# Test Plan

Successfully tested `./gradlew build` in `android`, as well as `react-native run-android` in `example`.

<sub><sup>Once approved, please use a *normal* GitHub merge (i.e. **NO rebase/squash** merge) to integrate the commit(s) from the PR head branch. The changes are broken up into meaningful, [atomic commits](https://www.freshconsulting.com/atomic-commits/), and my branch should already be up-to-date with the latest base branch. If it isn't, or if you want me to change anything, please let me know, and I will update the branch as soon as possible. Thank you!</sup></sub>

### DIFF
--- a/android/gradle/wrapper/gradle-wrapper.properties
+++ b/android/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.2-all.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
--- a/example/android/gradle/wrapper/gradle-wrapper.properties
+++ b/example/android/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.4.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.2-all.zip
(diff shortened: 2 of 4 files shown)


MateusAndrade:
Did you tested this on a production build? Also, this Works when using react-native-share with other libraries?

mikehardy:
Yeah these are basically rubber-stamp approvals, I don't fear gradle minors or point releases really. If it builds it should be fine

friederbluemle:
@MateusAndrade - Yes, good question. The Gradle wrapper only affects the project its part of. It does not get packaged and distributed for consumption by other apps.

mikehardy:
For completeness (and to check my understanding - I frequently learn things this way) - the gradle tool itself (currently at 5.x release series) is not packaged in any way with built code and does not even affect it. It is focused on running the build etc. Though I suppose gross errors like incorrect parallelization or build step ordering errors might result in bad builds I will ignore that for now. So those should be safe in general as changes

But the plugin (currently released with versions in lockstep with Android Studio) *can* affect built code as they include the dexer and shrinker (R8/D8) that emits the bytecode. So these need actual usage tests afterwards (though are in such wide use are in general safe as well, just maybe less so).

Is that understanding correct?

friederbluemle:
I think we were typing at the same time @mikehardy 😄 Your explanation is even more complete 👍

mikehardy:
Glad that looks right to you. And yes, I'm too wordy sorry :sweat_smile:

friederbluemle:
Rebased onto latest `master`. Please use a normal merge commit (also see comment in the PR description). Thanks!

MateusAndrade:
@mikehardy  Thanks clarifying the things about gradle. HAHAHAHA

Also, i just tested this with ```./gradlew build``` and ```react-native run-android``` with the example app, and it runs without any problems.
 
Thanks for helps us out @friederbluemle!

PR #584, opened 2019-09-13
Only call successCallback when completed is true (iOS 13 crash)

### LINKED ISSUE #583: Illegal callback invocation from native module. This callback type only permits a single invocation from native code
Crash with error `Illegal callback invocation from native module. This callback type only permits a single invocation from native code` when you cancel sharing

### Steps to reproduce
1. Use iOS 13
2. Open share dialog using Share.open
3. Tap message for example
4. Cancel message
5. Cancel share
6. Crash

### Expected behaviour
There should be no crash

### Actual behaviour
It crashed

### Environment
- **React Native version**: 0.60.5
- **React Native platform + platform version**: iOS 13.0

### react-native-share
**Version**: 2.0.0

https://github.com/

tomtargosz:
# Overview

Fixes #583.

Currently on iOS 13 the app will crash if you:
- Open the share sheet
- Tap something like messages or photos
- Cancel the dialog
- Perform any other action

This is because `shareController.completionWithItemsHandler` is called when the dialog box is canceled and currently `failureCallback` or `successCallback` will always be called. In the situation above, `activityError` is `nil` so `successCallback` will be called even though `completed` is false. This leaves us in a state where the callback has been invoked but the ShareSheet is still active, meaning the success or error callback will be invoked again, leading to the crash.

This PR adds a check to make sure `completed` is true before calling `successCallback`. This way `successCallback` will only be called when the user has successfully completed an action and the ShareSheet is closed.

# Test Plan

- Saved an image successfully
- Opened and dismissed the `Photos` dialog multiple times without crashing

### DIFF
--- a/ios/RNShare.m
+++ b/ios/RNShare.m
@@ -190,7 +190,7 @@ RCT_EXPORT_METHOD(open:(NSDictionary *)options
     shareController.completionWithItemsHandler = ^(NSString *activityType, BOOL completed, __unused NSArray *returnedItems, NSError *activityError) {
         if (activityError) {
             failureCallback(activityError);
-        } else {
+        } else if (completed) {
             successCallback(@[@(completed), RCTNullIfNil(activityType)]);
         }
     };


MateusAndrade:
Thanks for helping us @tomtargosz ! 🎉

PR #599, opened 2019-10-04
Remove unnecessary tools replace on build gradle

luancurti:
# Overview
Remove tools replace that was causing warnings

# Test Plan
Run build without problems

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -1,5 +1,4 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
     package="cl.json">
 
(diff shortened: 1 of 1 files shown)


luancurti:
Before this fix the warnings in build

Screen Shot 2019-10-03 at 23 43 21

mikehardy:
If I understand correctly this is might be related to the change where rnshare added a default provider etc? Could be considered a remnant in that case and good to clean out

MateusAndrade:
@luancurti did you tested this on the release and debug mode?

luancurti:
@MateusAndrade yes, before I made this PR I referenced my repository and installed in my app, running in debug and release mode, inside this project running `./gradlew assembleRelease` app build successfull

MateusAndrade:
Thanks @luancurti ! 🎉

PR #607, opened 2019-10-15
iOS 13: Resolve promise if ShareSheet is manually dismissed

### LINKED ISSUE #601: iOS 13 New UI close button doesn't reject promise
### Steps to reproduce
On iOS 13 we have issue with new close button:
If user clicks on new iOS 13 UI close button - .catch() callback doesn't fire, so Promise is not rejected in this case. On iOS 12 with previous UI Cancel button it works correctly

### Expected behaviour
```catch``` callback should be called, promise should be rejected.

### Actual behaviour
```catch``` callback is not called, promise is not rejected.

### Environment
- **React Native version**: 0.61.2
- **React Native platform + platform version**: iOS 13.0

### react-native-share
**Version**: 2.0.0"

tomtargosz:
# Overview

Fixes #601.

This PR fixes an issue on iOS 13 where the ShareSheet's promise isn't resolved if the user closes the sheet without completing any actions introduced in #584.

This PR changes the `if (completed)` check in `completionWithItemsHandler` to `if (completed || activityType == nil)`

It looks like in `completionWithItemsHandler` the activityType will only be `nil` if the user closed the ShareSheet themselves, like so:

Screen Shot 2019-10-14 at 1 41 22 PM

Here's when they close the reminders modal:
Screen Shot 2019-10-14 at 1 43 02 PM

And the photos modal:
Screen Shot 2019-10-14 at 1 47 03 PM

This should fix the current issue without reintroducing the crash.

# Test Plan

- Ensure promise is resolved when the ShareSheet is closed manually
- Ensure the promise is **not** resolved when a different modal is dismissed (photos, add to contact, reminders, etc)

### DIFF
--- a/ios/RNShare.m
+++ b/ios/RNShare.m
@@ -190,7 +190,7 @@ RCT_EXPORT_METHOD(open:(NSDictionary *)options
     shareController.completionWithItemsHandler = ^(NSString *activityType, BOOL completed, __unused NSArray *returnedItems, NSError *activityError) {
         if (activityError) {
             failureCallback(activityError);
-        } else if (completed) {
+        } else if (completed || activityType == nil) {
             successCallback(@[@(completed), RCTNullIfNil(activityType)]);
         }
     };


tomtargosz:
Similar PR is open in react-native: https://github.com/facebook/react-native/pull/26842

PR #619, opened 2019-10-30
fix: added missing "v" to `source` field in podspec

mdstroebel:
The tags are being named with a "v" and as such, the podspec needs to resolve the tags by including the "v" in the `source` field

# Overview

Without this fix, users are unable to link the library via CocoaPods. The Cocoapods install fails because it is unable to resolve the `source` of the pod.

# Test Plan


MateusAndrade:
This is related to auto-linking not working?

PR #633, opened 2019-11-15
Set useNativeDriver explicitly for Animations

James-E-Adams:
# Overview

Just noticed another flow issue when running flow in our codebase. Eventually useNativeDriver is a non-optional field https://github.com/facebook/react-native/commit/d123bea8c1a4f1c977928fa2cfffe19132f74dab
so may as well fix this up now.

# Test Plan

Run flow!


James-E-Adams:
I have a feeling the CI step failed because of flakiness, but can't rerun it myself.

James-E-Adams:
Yeah it is. I remember reading that eventually it will be true by default somewhere in the code.

We could make it a user decision with false by default - I wouldn't want it true by default because it can occasionally cause problems.

PR #700, opened 2020-02-02
Update podspec infos using package.json infos

luancurti:
# Overview

Avoid inconsistency of informations with `podspec` and `package.json`

# Test Plan

Go to `example/ios` run `pod install` and run example App and everything is working as expected


luancurti:
> nice @luancurti - we do the same in device-info, works well. You specified a different homepage here but I checked and it's valid and this way is better for this repo - +1

Yes the homepage in `package.json` has `#readme`. But I think is more clear to redirect to readme :)

PR #717, opened 2020-03-02
Fix share url android

### LINKED ISSUE #710: Share url in android. It is being shared between [ and " in version 3.0.0
### Steps to reproduce
1. Install version 3.0.0 of `react-native-share`
2. Share an url using Share.open (for example 'hola')
3. enjoy the bug

### Expected behaviour
The link has to be `hola`

### Actual behaviour
The link is `["hola"]``

If I downgrade the version to the 2.0.0 it works fine

luancurti:
# Overview

On sharing an url on android it shares the url inside an array

Fixes #710 

# Test Plan

Try to share an url and it shares correctly

### DIFF
--- a/android/src/main/java/cl/json/social/ShareIntent.java
+++ b/android/src/main/java/cl/json/social/ShareIntent.java
@@ -139,9 +139,9 @@ public abstract class ShareIntent {
                 }
             } else {
                 if (!TextUtils.isEmpty(message)) {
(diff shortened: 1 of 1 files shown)


luancurti:
@MateusAndrade can you publish the another version on npm with this fix?

MateusAndrade:
> @MateusAndrade can you publish a new version on npm with this fix?

Sure! I'm going to give a look on that this week.

PR #739, opened 2020-03-22
Update requirements to use xcode 11 and ios 13 sdk 

luancurti:
# Overview

Add xcode 11 to requirements and `LinkPresentation.h file not found` to `Troubleshooting` section

# Test Plan

Not applicable

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,11 @@ npm install react-native-share --save
 ```
 ---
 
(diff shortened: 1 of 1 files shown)


MateusAndrade:
Btw, I think our CI will deploy a new version, since it isn't looking for changes on docs only.

PR #750, opened 2020-04-06
fix excludeActivityTypes in android

rupam-ghosh:
This PR fixes excludeActivityTypes in Android N and above
We need to pass an array of ComponentName as value for key EXTRA_EXCLUDE_COMPONENTS but instead you are currently passing string.
Please check android documentation
https://developer.android.com/reference/android/content/Intent#EXTRA_EXCLUDE_COMPONENTS
Also check the android source code that reads this value
https://android.googlesource.com/platform/frameworks/base.git/+/master/core/java/com/android/internal/app/ChooserActivity.java#628


rupam-ghosh:
@mikehardy @jgcmarins @MateusAndrade Please review the PR

PR #754, opened 2020-04-07
Support sharing of files from secondary storage i.e sd cards, on Android

nirre7:
# Overview
Currently the Android config does not support sharing of files from secondary storage (sd cards). Developers/users have used the base64 workaround for this, but I thought it would be a good idea to support it.
Also none of these would solve it:

```
  <external-files-path
    name="external_files"
    path="." />
  <external-cache-path
    name="external_cache"
    path="." />
<files-path
    name="files"
    path="." />

```
# Test Plan
On an emulator or device, download an image to the sdcard or "external storage" and then share it.

### DIFF
--- a/android/src/main/java/cl/json/RNSharePathUtil.java
+++ b/android/src/main/java/cl/json/RNSharePathUtil.java
@@ -57,7 +57,7 @@ public class RNSharePathUtil {
                     break;
                 }
             } catch (Exception e) {
-
+                System.out.println("ERROR " + e.getMessage());
             }
         }
         return result;
--- a/android/src/main/java/cl/json/RNSharePathUtil.java
+++ b/android/src/main/java/cl/json/RNSharePathUtil.java
@@ -57,7 +57,7 @@ public class RNSharePathUtil {
                     break;
                 }
             } catch (Exception e) {
-                System.out.println("ERROR " + e.getMessage());
+                System.out.println("RNSharePathUtil::compatUriFromFile ERROR " + e.getMessage());
             }
         }
         return result;
--- a/android/src/main/res/xml/share_download_paths.xml
+++ b/android/src/main/res/xml/share_download_paths.xml
@@ -2,4 +2,5 @@
 <paths xmlns:android="http://schemas.android.com/apk/res/android">
     <external-path name="rnshare1" path="Download/" />
     <cache-path name="rnshare2" path="/" />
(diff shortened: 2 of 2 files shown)


MateusAndrade:
Hi @nirre7 , thanks for that! Would you mind applying these changes to the example app? Also, this shouldn't be a problem setting this key `<root-path name="rnshare_sdcard" path="." />` to `rnshare_sdcard`?

android/src/main/java/cl/json/RNSharePathUtil.java
```diff
@@ -57,7 +57,7 @@ public static Uri compatUriFromFile(@NonNull final ReactContext context, @NonNul
                     break;
                 }
             } catch (Exception e) {
-
+                System.out.println("ERROR " + e.getMessage());
```
mikehardy, line 60: This would be better with some prefix indicating where it came from, e.g. 
```suggestion
                System.out.println("RNSharePathUtil::compatUriFromFile ERROR " + e.getMessage());
```

MateusAndrade: > This would be better with some prefix indicating where it came from, e.g. 
> ```suggestion
>                 System.out.println("RNSharePathUtil::compatUriFromFile ERROR " + e.getMessage());
> ```

Agree with that.

nirre7:
> rnshare_sdcard

Afaik the name attribute "rnshare_sdcard" could be whatever as long as it is not the same as other path configs. i.e the following wouldn't work: 
```
    <cache-path name="test" path="." />
    <root-path name="test" path="." />
```

nirre7:
Also thank you for reviewing this so swiftly!

nirre7:
>  Would you mind applying these changes to the example app?
Not sure what you want me to do? 
- There is nothing that needs to be done in an app that is using this lib, from a configuration perspective.
- The example app is using a base64 image (since it is "portable") for it's examples which gets handled by a different code path in the lib. 

It might be me which is misunderstanding though 🤔

MateusAndrade:
> It might be me which is misunderstanding though 🤔

I think i misunderstand. HAHAHA Sorry for that.

nirre7:
> > It might be me which is misunderstanding though 🤔
> 
> I think i misunderstand. HAHAHA Sorry for that.

Ok, td bem 👍

Issue #676, opened 2020-01-16
App Icon is missing


Sach-SPF:
### Steps to reproduce
1. iOS 13
2. Add "message" property to the options
3. App icon is missing

### Expected behaviour
App Icon should appear

### Actual behaviour
App Icon is missing

### Environment
- **React Native version**: 59.9

The app icon seems to be hidden or remove or missing when I used "Message" property else its appear

ActionSheet without Message 
<img width="260" alt="Screen Shot 2020-01-16 at 7 33 28 PM" src="https://user-images.githubusercontent.com/28600651/72531356-465df000-3897-11ea-8f3e-d7ecb416a4bc.png">

ActionSheet with Message

<img width="264" alt="Screen Shot 2020-01-16 at 7 37 09 PM" src="https://user-images.githubusercontent.com/28600651/72531466-8e7d1280-3897-11ea-9f61-20cc9338e1e3.png">

<img width="358" alt="Screen Shot 2020-01-16 at 7 38 01 PM" src="https://user-images.githubusercontent.com/28600651/72531531-ac4a7780-3897-11ea-9250-3da3751a0a42.png">

Thanks
Sachin

RavindraV07:
Iam facing same issue in iOS V13.3

ansalspurtree:
I am also facing the same issue. any fix?

bataevvlad:
Facing same issue

sercanuste:
Hi,

You can use following code for App Icon.

Tested on; iOS 13.3.1,
"react": "16.8.3",
"react-native": "0.59.9",
"react-native-share": "3.1.2"

---
```javascript
const url = 'https://awesome.contents.com/';
const message = 'Please check this out.';
const logo = 'data:<data_type>/<file_extension>;base64,<base64_data>';

const options = Platform.select({
    ios: {
        activityItemSources: [
            {
                placeholderItem: {
                    type: 'url',
                    content: logo
                },
                item: {
                    default: {
                        type: 'text',
                        content: `${message} ${url}`
                    },
                },
                linkMetadata: {
                    title: message,
                    icon: logo
                }
            }
        ],
    },
    default: {
        message: `${message} ${url}`
    },
});

Share.open(options)
    .then((res) => { console.log(res) })
    .catch((err) => { console.log(err); });
```

sumit31agarwal:
what about if we are using expo ,  I am using Share from react native and below is the screenshot what I am getting 
[...]

PR #756, opened 2020-04-08
docs: Custom icon usage instructions for iOS share preview added to example

sercanuste:
Example for: #676

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,10 +223,11 @@ import Share from 'react-native-share';
 const url = 'https://awesome.contents.com/';
 const title = 'Awesome Contents';
 const message = 'Please check this out.';
+const icon = 'data:<data_type>/<file_extension>;base64,<base64_data>';
 const options = Platform.select({
   ios: {
     activityItemSources: [
-      {
+      { // For sharing url with custom title.
         placeholderItem: { type: 'url', content: url },
         item: {
           default: { type: 'url', content: url },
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -223,10 +223,11 @@ import Share from 'react-native-share';
 const url = 'https://awesome.contents.com/';
 const title = 'Awesome Contents';
 const message = 'Please check this out.';
+const icon = 'data:<data_type>/<file_extension>;base64,<base64_data>';
```
MateusAndrade, line 226: Only works with base64?

MateusAndrade:
Hey @sercanuste , would you mind showing an example of that on iOS? I haven't tested the new share system on iOS very often. This would be nice to have an image example, even if juts on this PR. 🤗

README.md
sercanuste: I tested with url. It overrides link metadata when image from url loaded.

sercanuste:
Hi,

First image is the default behavior with message 
```jsx
{
  message: `${message} ${url}`
}
```
image

And this is with activityItemSources
```jsx
{
    activityItemSources: [
        {
            placeholderItem: {
                type: 'url',
                content: icon
            },
            item: {
                default: {
                    type: 'text',
                    content: `${message} ${url}`
                },
            },
            linkMetadata: {
                title: message,
                icon: icon
            }
        }
    ],
}
```
image

sercanuste:
I found out while testing that using activityItemSources like this removes default text icon and shows app icon on share preview. This can be real answer for #676 and PR is for custom icon alternative.

```jsx
activityItemSources: [
    {
        placeholderItem: {
            type: 'text',
            content: `${message} ${url}`
        },
        item: {
            default: {
                type: 'text',
                content: `${message} ${url}`
            },
        },
        linkMetadata: {
            title: message,
        }
    }
],
```

README.md
MateusAndrade: Maybe we can add something like:

```suggestion
const icon = 'data:<data_type>/<file_extension>;base64,<base64_data>' or "remote-image-url";
```

This should work?

sercanuste: When I use remote url it loads title and url from internet and overrides linkMetadata.
```jsx
const icon = 'https://www.sercanuste.com/img/logo/BA.png';
```
IMG_162F379A8183-1

PR #761, opened 2020-04-13
feat: Implement Facebook stories share

Quynh-Nguyen:
# Overview
Implement Facebook Stories share.
Follow this document of Facebook: https://developers.facebook.com/docs/sharing/sharing-to-stories/ios-developers

# Test Plan
- Add more handler for share to Facebook.
- Tested with my React Native project on iPhone 7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -535,7 +535,7 @@ import Share from 'react-native-share';
 
 const shareOptions = {
     method: Share.FacebookStories.SHARE_BACKGROUND_AND_STICKER_IMAGE,
-    backgroundImage: 'http://urlto.png',
+    backgroundImage: 'http://urlto.png', // url or an base64 string
     stickerImage: 'data:image/png;base64,<imageInBase64>', //or you can use "data:" link
     backgroundBottomColor: '#fefefe',
     backgroundTopColor: '#906df4',
--- a/README.md
+++ b/README.md
@@ -17,12 +17,12 @@ This gives you the power to prioritize our work and support the project contribu
 
 ## Automatic Way
 ---
-``` 
+```
 yarn add react-native-share
 ```
 
 or if you're using npm
-``` 
+```
 npm install react-native-share --save
 ```
 ---
@@ -35,7 +35,7 @@ npm install react-native-share --save
 #### Important:
 Linking is not needed anymore. ``react-native@0.60.0+`` supports dependencies auto linking.
 For iOS you also need additional step to install auto linked Pods (Cocoapods should be installed):
-``` 
+```
 cd ios && pod install && cd ../
 ```
 ___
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -526,10 +527,49 @@ Supported options for INSTAGRAM_STORIES:
...
 
+### Static Values for Facebook Stories
+These can be assessed using Share.Social property
+For eg.
+```javascript
+import Share from 'react-native-share';
+
+const shareOptions = {
+    method: Share.FacebookStories.SHARE_BACKGROUND_AND_STICKER_IMAGE,
+    backgroundImage: 'http://urlto.png',
```
MateusAndrade, line 538: Does this accept a `base64`?

README.md
```diff
@@ -526,10 +527,49 @@ Supported options for INSTAGRAM_STORIES:
...
+    stickerImage: 'data:image/png;base64,<imageInBase64>', //or you can use "data:" link
```
MateusAndrade, line 539: Do this works sharing with a URL?

README.md
```diff
@@ -526,10 +527,49 @@ Supported options for INSTAGRAM_STORIES:
...
+    backgroundBottomColor: '#fefefe',
+    backgroundTopColor: '#906df4',
+    attributionURL: 'http://deep-link-to-app', //in beta
+    appId: '219376304', //facebook appId
```
MateusAndrade, line 543: This is the appId on Facebook Pages?

index.js
```diff
@@ -92,6 +92,7 @@ type Options = {
   failOnCancel?: boolean,
   showAppsToView?: boolean,
   saveToFiles?: boolean,
+  appId?: string,
```
MateusAndrade, line 95:
```suggestion
  appId: string,
```

This should be required, no?

README.md
Quynh-Nguyen: Yes it's working both with base64 data image and url.

README.md
Quynh-Nguyen: Yes it's working with both base64 data image and url.

README.md
Quynh-Nguyen: Yes that's right, it's required from Facebook.
With Instagram we doesn't need, but with Facebook they required.

README.md
MateusAndrade: Maybe we can let implicit  that this param accept both `url` and `base64`.

README.md
MateusAndrade: Maybe we can let implicit  that this param accept both `url` and `base64`.

Quynh-Nguyen:
> Also, could you provide an example to test this? Since we're using Semantic Release, whenever this gets merged a new version will be published, so we need to be careful. 😅
> 
> I will try testing this here and after that, we can merge this. 🎉

Yah this is the testing result.

image

README.md
Quynh-Nguyen: > Maybe we can let implicit that this param accept both `url` and `base64`.
Yes. So should me update the example source code on README file?

README.md
Quynh-Nguyen: > Maybe we can let implicit that this param accept both `url` and `base64`.

Yes. So should me update the example source code on README file?

README.md
MateusAndrade: Yeah, sure! I approved this before seeing your comment. 😅

README.md
MateusAndrade: Same here.

README.md
MateusAndrade, line 538:
```suggestion
    backgroundImage: 'http://urlto.png', // url or an base64 string
```

README.md
MateusAndrade, line 539:
```suggestion
    stickerImage: 'data:image/png;base64,<imageInBase64>', //or you can use "data:" url
```

PR #770, opened 2020-04-25
Only convert options.url to options.urls in iOS

i1skn:
# Overview
In Android this behaviour would result in `Intent.ACTION_SEND_MULTIPLE` instead of `Intent.ACTION_SEND` for a single file. Some apps could support one, but not another.

# Test Plan
I've checked in on a real device using https://f-droid.org/en/packages/de.k3b.android.intentintercept/


MateusAndrade:
Thanks!

PR #773, opened 2020-05-05
Security alerts / package updates 20200505

mikehardy:
# Overview

I get security alerts for this repo and I let them sit for a couple weeks but finally got irritated at receiving the same one multiple times :-)

This is a series of commits:
1. basic upgrade of the vulnerable packages by just deleting yarn.lock and re-running yarn
1. some atomic commits that fix up all that's needed in order to pass `yarn validate` if you...
1. `yarn upgrade --latest` (with the exception of flow-bin, it has to match current react-native)

All seems fine?

# Test Plan

yarn validate


mikehardy:
Your devious strategy worked!! 😆  no worries, it's open source, if it bugs me, I should fix it, and here we are. Cheers man

PR #774, opened 2020-05-05
Recommend pod-install for setup instructions

EvanBacon:
We've been recommending devs use `npx pod-install` since it will attempt to install CocoaPods CLI if it's not available on the computer (cite [React Navigation setup guide](https://reactnavigation.org/docs/getting-started/#installing-dependencies-into-a-bare-react-native-project)). This has proved very useful for Expo users who are now migrating to the bare workflow and want to use community packages in their projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,10 +33,11 @@ npm install react-native-share --save
 2. iOS 13 SDK or higher
 
 #### Important:
+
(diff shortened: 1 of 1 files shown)


EvanBacon:
Ya! It'll also run `pod repo update` if the command fails, then attempt to pod install again. It has definitely helped a lot of beginners so far!

PR #813, opened 2020-06-28
feat: adding recipient for sms on Android

### LINKED ISSUE #812: Add recipients for SMS sharing on Android
## Describe the Feature
Be able to specify recipients when sending SMS on Android. 

## Possible Implementations
When using SMS, have a field in the options object that specifies the recipients as an array of phone numbers, for example  `{ ... social: 'sms', recipients: Array<string> ...}`.

## Related Issues
#762

Ackuq:
# Overview
Add an optional field `recipient` for SMS destination recipient, resolves #812.

When choosing to share using SMS, you can include one recipient as a string. For example: 
```js
Share.shareSingle({
      title: 'Hello SMS!',
      social: Share.Social.SMS,
      recipient: '123-4567890',
      message: 'Example SMS message',
});
```

### But why not multiple recipients?

Different vendors have different formatting when specifying multiple destination phone numbers. And also, some vendors do not even support adding multiple destination phone numbers when using the `ACTION_SEND` intent. Which would require the `ACTION_SENDTO` intent. However, could be implemented in the future.

# Test Plan
Test it through the example app.

recipient-example

### DIFF
--- a/android/src/main/java/cl/json/social/ShareIntent.java
+++ b/android/src/main/java/cl/json/social/ShareIntent.java
@@ -121,18 +121,10 @@ public abstract class ShareIntent {
         }
 
         if (socialType.equals("sms")) {
-            String[] recipientsArray = options.getArray("recipients");
-            String separator = ",";
-            if(android.os.Build.MANUFACTURER.equals("Samsung")){
-                separator = ";";
-            }
-            String recipients = "";
-            for (int i = 0; i < recipientsArray.size(); i++) {
-                recipients += recipientsArray.getString(i);
-                recipients += separator;
-            }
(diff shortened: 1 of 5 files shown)


Ackuq:
When using `Share.open({...})` this feature will only work if setting the social attribute to `'sms'` or `'Share.Social.SMS'`. Because of the `if (socialType.equals("sms")) { ... }` inside of the `ShareIntent.java`. But maybe it is better to add it even though the social attribute not being set to 'sms'? Sort of like how the `email` attribute is being handled right now. So the aforementioned statement would be replaced with: `if (ShareIntent.hasValidKey("recipient", options)) { ... }`

Issue #805, opened 2020-06-17
Port project to Typescript + Flow Types 


MateusAndrade:
## Describe the Feature
Since we have been doing a lot of changes to `react-native-share` with new documentation and improvements on the  `example` application, aiming to make the life of any contributor easier. 

Following the example of [react-navigation](https://github.com/react-navigation/react-navigation) which mainly uses TS now, it would simpler to update types, etc.

My idea is porting the main files which are currently written in plain `.js` with `flow` to `ts`. But maintaining the support for flow types. (Any examples are welcome) 
  

## Possible Implementations

Firstly, we need to map the common APIs exported from `react-native-share` and their respective/expected types. We can use the types from [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped#readme) as a start point.

After that, we only need to ensure that the compiled version of these files is working with both `ts`, `flow`, and `js`  projects without any drawbacks.

## Roadmap

- [ ] Port main js files to TS
- [ ] Ensure that both `flow` types and `ts` are supported
- [ ] Compile project on CI before publish to `npm`

mikehardy:
Another project I check went further without incident - just convert to typescript, evict the flow types and if people want flow-typing they can contribute types to flow-typed repo. It's basically reversing the previous stance (flow first / typescript secondary) but in a way that - while opinionated - recognizes the current javascript ecosystem reality. Typescript won.

jgcmarins:
That's great!
Check how [graphql-js has been doing that](https://github.com/graphql/graphql-js/issues/2104).

Thomazella:
👀 I can help with Ts but not so much with flow. I think the decision to go full TS is simpler than maintaining both TS and Flow. I'd still keep a branch with the flow types, just to have them as reference.

MateusAndrade:
> 👀 I can help with Ts but not so much with flow. I think the decision to go full TS is simpler than maintaining both TS and Flow. I'd still keep a branch with the flow types, just to have them as reference.

Sure! I'm a newbie with flow, maybe @jgcmarins can help us. I think I will push a new branch as soon as we start working on that to become our "backup". Do you have any thoughts about removing the types from DefinetlyTyped or something similar to that?

MateusAndrade:
> > 👀 I can help with Ts but not so much with flow. I think the decision to go full TS is simpler than maintaining both TS and Flow. I'd still keep a branch with the flow types, just to have them as reference.
> 
[...]

PR #818, opened 2020-07-05
Feat/ts

tcodes0:
# Overview
see #805 
Introduces typescript to project

# Test Plan
run `yarn validate:ts`

# todo

- [x] deps and config
- [x] components/
- [x] eslint-typescript configuration, should fix ci
- [ ] figure out how to build ts for npm
- [x] type index.js (maybe in another pr)
- [x] flow: remove?
- [ ] read bridge and double check types


Thomazella:
Do we want TS to emit types and declarations or just type-check?

components/Button.tsx
```diff
@@ -0,0 +1,52 @@
...
+  icon: {
+    width: 28,
+    height: 28,
+    marginLeft: 10,
+    marginRight: 30,
+  },
+});
+
+type Props = {
+  buttonStyle?: StyleProp<ViewStyle>;
```
Thomazella, line 37: is this optional?

components/Button.tsx
```diff
@@ -0,0 +1,52 @@
...
+  onPress: (event?: GestureResponderEvent) => void;
+  iconSrc: ImageSourcePropType;
+  textStyle?: StyleProp<TextStyle>;
```
Thomazella, line 40: is this optional?

package.json
```diff
@@ -68,8 +72,7 @@
   "license": "MIT",
   "lint-staged": {
     "*.js": [
-      "yarn validate",
-      "git add"
```
Thomazella, line 72: cli had a warning here so I fixed it

mikehardy:
Ohhh :drooling_face: 
Your question on emitting types and declarations: I think one of the stated goals of the conversion is to be the source of truth for typing (so that people don't need to import from definitely typed). In order to do that we must necessarily emit the types and declarations I think?

.eslintrc
```diff
@@ -1,45 +1,79 @@
...
-          ".json"
-        ]
+        "extensions": [".js", ".android.js", ".ios.js", ".json", ".ts", ".tsx"]
       }
     }
   },
   "globals": {
     "__DEV__": true
+  },
+  "rules": {
```
Thomazella, line 33: just rules I use

package.json
```diff
@@ -78,15 +83,16 @@
     "start": "react-native start",
     "start:android": "react-native run-android",
     "start:ios": "react-native run-ios",
-    "lint": "eslint index.js --max-warnings=0",
```
Thomazella, line 81: why allow no warnings?

Thomazella:
@MateusAndrade on a good spot for a review, specially around the ts-ignore stuffs

index.tsx
```diff
@@ -0,0 +1,346 @@
...
+        PermissionsAndroid.PERMISSIONS.WRITE_EXTERNAL_STORAGE,
+      );
+      if (hasPermission) {
+        return Promise.resolve(true);
+      }
+      const result = await PermissionsAndroid.request(
+        PermissionsAndroid.PERMISSIONS.WRITE_EXTERNAL_STORAGE,
+      );
+      if (result === PermissionsAndroid.RESULTS.GRANTED) {
+        return Promise.resolve(true);
```
Thomazella, line 206: added `true` here was undefined

index.tsx
```diff
@@ -0,0 +1,346 @@
...
+  placeholderItem: ActivityItem;
+  item: { markupAsPDF: ActivityItem } | { [key: string]: ActivityItem };
+  subject?: Partial<IActivityType> | { [key: string]: string };
+  dataTypeIdentifier?: Partial<IActivityType> | { [key: string]: string };
+  thumbnailImage?: Partial<IActivityType> | { [key: string]: string };
+  linkMetadata?: LinkMetadata;
+};
+
+type OpenReturn = { app?: string; dismissedAction?: boolean };
+type AppMessageReturn = { app: ActivityType; message: ActivityType };
```
Thomazella, line 168: new type

index.tsx
```diff
@@ -0,0 +1,346 @@
...
+    } catch (e) {
+      return Promise.reject(e);
+    }
+  }
+  return Promise.resolve(true);
+};
+
+class RNShare {
+  static Button: any;
+  static ShareSheet: any;
```
Thomazella, line 218: added `any` here

index.tsx
```diff
@@ -0,0 +1,346 @@
...
+    if (this.props.visible) {
+      this.props.onCancel();
+      return true;
+    }
+    return false;
+  }
+  render() {
+    const { style = {}, overlayStyle = {}, visible, ...props } = this.props;
+    return (
+      <Overlay visible={visible} {...props}>
```
Thomazella, line 65: spread was overwriting prop always

index.tsx
```diff
@@ -0,0 +1,346 @@
...
+  title?: string;
+  subject?: string;
+  activityItemSources?: Array<ActivityItemSource>;
+  excludedActivityTypes?: string;
+  failOnCancel?: boolean;
+  showAppsToView?: boolean;
+  saveToFiles?: boolean;
+};
+
+enum ActivityType {
```
Thomazella, line 109: new enum

index.tsx
```diff
@@ -0,0 +1,346 @@
...
+  remoteVideoUrl?: string;
+  video?: string;
+};
+
+type ActivityItemSource = {
+  placeholderItem: ActivityItem;
+  item: { markupAsPDF: ActivityItem } | { [key: string]: ActivityItem };
+  subject?: Partial<IActivityType> | { [key: string]: string };
+  dataTypeIdentifier?: Partial<IActivityType> | { [key: string]: string };
+  thumbnailImage?: Partial<IActivityType> | { [key: string]: string };
```
Thomazella, line 163: lmk about this typing here

index.tsx
```diff
@@ -0,0 +1,346 @@
...
+  postToFlickr = 'postToFlickr',
+  postToTencentWeibo = 'postToTencentWeibo',
+  postToTwitter = 'postToTwitter',
+  postToVimeo = 'postToVimeo',
+  postToWeibo = 'postToWeibo',
+  print = 'print',
+  saveToCameraRoll = 'saveToCameraRoll',
+  markupAsPDF = 'markupAsPDF',
+}
+interface IActivityType {
```
Thomazella, line 127: new interface

package.json
mikehardy: I don't intent to set policy but in my experience lint is like - either you do it all the way, or you don't
I usually set all my projects like this, then for very specific cases you can lint ignore a specific line in a file, with justification in the associated PR. Too harsh? I dunno, works in practice though

Thomazella: I agree, I think it's better to keep the flag as it was before, we can ignore inline

package.json
```diff
@@ -78,15 +83,16 @@
...
+    "lint": "eslint index.tsx",
     "lint:staged": "lint-staged",
     "release:major": "node ./changelog --major && npm version major && git push origin && git push origin --follow-tags",
     "release:minor": "node ./changelog --minor && npm version minor && git push origin && git push origin --follow-tags",
     "release:patch": "node ./changelog --patch && npm version patch && git push origin && git push origin --follow-tags",
     "release:version": "node ./changelog --version $VERSION && git push origin && git push origin --follow-tags",
-    "validate": "yarn validate:prettier && yarn validate:lint && yarn validate:flow",
-    "validate:flow": "yarn flow",
-    "validate:lint": "eslint *js components/*.js --max-warnings=0",
+    "validate": "yarn validate:prettier && yarn validate:lint && yarn validate:flow && validate:ts",
```
luancurti, line 92: I think you can remove flow validation

```suggestion
    "validate": "yarn validate:prettier && yarn validate:lint && yarn  validate:ts",
```

components/Button.tsx
MateusAndrade: Yes

components/Button.tsx
MateusAndrade: Yes

components/Button.tsx
```diff
@@ -0,0 +1,52 @@
...
+};
+
+const Button: React.FC<Props> = ({ buttonStyle, onPress, iconSrc, textStyle, children }) => {
```
MateusAndrade, line 43:
```suggestion
const Button: React.FC<Props> = ({ buttonStyle = {}, onPress, iconSrc, textStyle = {}, children }) => {
```

Since both are optional I prefer to set a default value to these props instead of passing an `undefined` directly to the component.

components/Button.tsx
```diff
@@ -0,0 +1,52 @@
...
+  },
+  icon: {
+    width: 28,
+    height: 28,
+    marginLeft: 10,
+    marginRight: 30,
+  },
+});
+
+type Props = {
```
MateusAndrade, line 36: Also, I think it would be a good idea to export the types from this component since it would be a lot easier to test use their props out of the library instead of rewriting these props again.

What do you think?

index.tsx
MateusAndrade: Do you think It would be better to avoid the `props` to override the `visible`?

```suggestion
      <Overlay  {...props} visible={visible}>
```

index.tsx
MateusAndrade: Something like: `React.FC<Props>` would work?

index.tsx
```diff
@@ -0,0 +1,346 @@
...
+    TWITTER: NativeModules.RNShare.TWITTER || 'twitter',
+    WHATSAPP: NativeModules.RNShare.WHATSAPP || 'whatsapp',
+    INSTAGRAM: NativeModules.RNShare.INSTAGRAM || 'instagram',
+    INSTAGRAM_STORIES: NativeModules.RNShare.INSTAGRAM_STORIES || 'instagramstories',
+    GOOGLEPLUS: NativeModules.RNShare.GOOGLEPLUS || 'googleplus',
+    EMAIL: NativeModules.RNShare.EMAIL || 'email',
+    PINTEREST: NativeModules.RNShare.PINTEREST || 'pinterest',
+    LINKEDIN: NativeModules.RNShare.LINKEDIN || 'linkedin',
+    SMS: NativeModules.RNShare.SMS || 'sms',
+  };
```
MateusAndrade, line 234: What do you think about a `enum` to reference these values instead of hardcoded strings?

Thomazella:
I'm hoping we can emit that file using the source code as base.

mikehardy:
Yep - device-info emits the file for what it's worth, using `bob` as a build tool it just happens?
I'm not really watching this closely so nothing just wanted to say in general, nice work :-), I'm excited to see the result

index.tsx
Thomazella: ts reported it as an error, I'd rather not touch any logic tho, so yeah, will try to go back to overwriting it

components/Button.tsx
Thomazella: yep, that works. I think it's fine with undefined tho

components/Button.tsx
Thomazella: definitely

Thomazella:
So the last thing is to build the js with tsc, and double check types. I think we'll get the .d.ts files "for free" like that, which should make it easier to maintain. I think it may make sense to type the bridge stuff too, but we'd probably need to do that in a file and maintain it. Maybe another PR for that. Overall we're close.

mikehardy:
All of that should be for free with bob I think here is the stanza from device info https://github.com/react-native-community/react-native-device-info/blob/master/package.json

```json
  "@react-native-community/bob": {
    "source": "src",
    "output": "lib",
    "targets": [
      "commonjs",
      "module",
      "typescript"
    ]
  }
```

Thomazella:
@mikehardy ok, didn't know bob, I get it now. Thanks!

index.tsx
AntoineDoubovetzky: @Thomazella what about something like
```
subject?: { [key in ActivityType]?: string };
```
instead of 
```
subject?: Partial<IActivityType> | { [key: string]: string };
```
?

This way you don't need the `IActivityType` interface anymore :)

Thomazella: Looks better, thanks! I have to finish this PR.

luancurti:
@Thomazella what's the status of this PR? Do you need some help?

Thomazella:
@luancurti nope, should be able to finish, will aim for this week. I was just busy with my new job that's all.

MateusAndrade:
> @luancurti nope, should be able to finish, will aim for this week. I was just busy with my new job that's all.

Let me know if you need any help. 👯

DanielMarkiel:
@MateusAndrade , @tcodes0 
What's the status of this PR? Is it stale (nothing has changed since Jul 13, 2020)?
If yes, would you mind if I try completing it next week?

tcodes0:
feel free to complete if :)

MateusAndrade:
> @MateusAndrade , @tcodes0
> What's the status of this PR? Is it stale (nothing has changed since Jul 13, 2020)?
> If yes, would you mind if I try completing it next week?

Let me know if there is anything we can do to help. 🚀

MateusAndrade:
Thanks a lot, guys, we merged #1001 and successfully published it under the version [v6.0.0](https://github.com/react-native-share/react-native-share/releases/tag/v6.0.0). 🚀

PR #575, opened 2019-08-31
Sharing to snapchat on iOS through CreativeKit

VincentRoest:
# Overview

Well, sharing to Snapchat was a key necessity for my app, so I decided I'd try my luck. Sadly, we have to depend on SnapSDK's CreativeKit. This also requires that production apps need to be reviewed. 

I realised my iOS development skills were a bit rusty, and that Snap's documentation is not ideal during the process. Nonetheless, I got this hacky version working for sharing videos, images, stickers and text content to Snapchat from iOS. The code is not very clean yet, and we could definitely improve it. Especially the "sleep" could use improvement, but I couldn't find a clean way of preventing the `_scSdkSnapApi` to be released before the share was complete during my late night struggle. All suggestions/edits are very welcome.

# Test Plan

1. Create a developer account for Snapchat here https://kit.snapchat.com/portal/, and create a new app there. 
2. Select Creative kit as a kit you will use in your app
3. Add (or modify) the following keys to your project's Info.plist **SCSDKClientId** (I put in the value for my app's "CLIENT ID FOR NON-CONFIDENTIAL OAUTH2 FLOW"), **LSApplicationQueriesSchemes** where you add 'snapchat' in the array just like with whatsapp or facebook. Optional: **SCSDKRedirectUrl**, some redirect URL, like: my-app://. If you add this, **CFBundleURLSchemes** must also contain your redirect URL’s scheme. More info here: https://docs.snapchat.com/docs/creative-kit
4. add `pod 'SnapSDK', :subspecs => ['SCSDKCreativeKit']` to your project's PodFile. @ _maintainers_, could this be optionally included inside the react-native-share project?

# Example code:
- I'm using the "title" options property as the Snap's text component
- I'm using the "attachmentUrl" for attaching a url to the snap.
```
import Share, {ShareSheet}  from 'react-native-share';
import RNFetchBlob from 'rn-fetch-blob'

shareToSnapchat = async (
    mediaUrl = "https://sample-videos.com/video123/mp4/480/big_buck_bunny_480p_1mb.mp4",
  ) => {
    try {
      RNFetchBlob.config({
        fileCache : true,
        appendExt : 'mp4'
      }).fetch('GET', mediaUrl, {})
      .then((res) => {
        let shareOptions = {
          title: "SNAP TEXT CONTENT",
          url: 'file://' + res.path(),
          type: 'video/mp4', // change type here accordingly
          attachmentUrl: 'https://snapchat.com',
          social: "snapchat",
        }
[...]

PR #825, opened 2020-07-16
Removed problematic parts of #575

NishidaRyu416:
# Overview

# Test Plan


ios/InstagramStories.m
```diff
@@ -22,7 +22,7 @@ @implementation InstagramStories
 
 - (void)backgroundImage:(NSData *)backgroundImage attributionURL:(NSString *)attributionURL {
     // Verify app can open custom URL scheme, open if able
-    
+
```
mikehardy, line 25: Nothing in this file changed except whitespace, which unnecessarily slows down review

ios/RNShare.m
```diff
@@ -47,6 +47,12 @@
 #import "FacebookStories.h"
 #import "GooglePlusShare.h"
 #import "EmailShare.h"
+
+
+#import "InstagramStories.h"
+// add snapchatshare
```
mikehardy, line 53: strange whitespace here and the comment seems unnecessary - all the imports should just go one after the other I think?

ios/SnapchatShare.h
```diff
@@ -0,0 +1,52 @@
+//
+//  GooglePlusShare.h
+//  RNShare
+//
+//  Created by Diseño Uno BBCL on 23-07-16.
+//  Copyright © 2016 Facebook. All rights reserved.
```
mikehardy, line 6: These aren't right, should be copyright you with license MIT or similar

ios/SnapchatShare.h
```diff
@@ -0,0 +1,52 @@
...
+//
+#import <UIKit/UIKit.h>
+// import RCTConvert
+#if __has_include(<React/RCTConvert.h>)
+#import <React/RCTConvert.h>
+#elif __has_include("RCTConvert.h")
+#import "RCTConvert.h"
+#else
+#import "React/RCTConvert.h"   // Required when used as a Pod in a Swift project
+#endif
```
mikehardy, line 16: I'm not sure if these are still necessary in 3 different ways? Is it possible to just import directly nowadays?

ios/SnapchatShare.m
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+    return self;
+}
+
+- (void *)shareSingle:(NSDictionary *)options
+    failureCallback:(RCTResponseErrorBlock)failureCallback
+    successCallback:(RCTResponseSenderBlock)successCallback {
+
+    NSLog(@"Try open view");
```
mikehardy, line 36: I'm not sure if other share options directly log but I don't think that's good practice? At minimum if you do log it should contain enough information to know exactly where to locate the log for problem localization, e.g. 'RNShare::shareSingle - Try open view'

package-lock.json
```diff
@@ -0,0 +1,15982 @@
+{
```
mikehardy, line 1: I don't believe this file should be included as part of the change?

NishidaRyu416:
> My biggest question on this one is if the Snapchat SDK could be made optional? I'm not sure what this will do to app size for people that don't use Snapchat? I believe for instance facebook share works but without directly including the facebook SDK so I'm not sure why the Snapchat SDK has to be brought in as a direct dependency

We need the SDK because all other sharing has been done with deep linking, which you can grasp by seeing the implementations, however SnapChat needs it because it does not support sharing with deep linking.

package-lock.json
NishidaRyu416: Maybe #575's author is using npm other than yarn?
Should I include a line to exclude it to .gitignore for future comers?

mikehardy:
Ah - I didn't realize you were picking up the baton from another PR, yeah, that is a thankless job :rofl: - I'm still not sure about the headers being tried in 3 ways - is that really needed? - but cool the SDK can be optional. Not sure if it will still compile correctly if it is optional, the whole implementation might need an IFDEF check for the Snapchat SDK header definition then - but still might be possible

NishidaRyu416:
@mikehardy, where to install modules specifically for the example app?
It seems that rn-fetch-blob is needed

mikehardy:
that would be in example/package.json or example/ios/Podfile I think?

NishidaRyu416:
https://react-native-community.github.io/react-native-share/docs/contributing#where-is-the-packagejson-and-node_modules

and after taking a look at https://github.com/react-native-community/datetimepicker/blob/488bedac6a41ccb8c6b34d2a07824de6fd77f47a/package.json, I guess we should put it in devDependencies.

NishidaRyu416:
Unfortunately, I cannot get the example app worked: snapchat does not launch although I follow the doc.
Maybe the SDK is broken?
I hope somebody who is more familiar with native codes investigate this more:)

NishidaRyu416:
Snap Inc. says they have fixed the SDK so the code might work as expected after updating the SDK!

franzejr:
@NishidaRyu416 what is the version you are currently using ?

Issue #846, opened 2020-08-24
shareSingle: email and attachments not working


cristianoccazinsp:
### Steps to reproduce
1. Call `shareSingle` with the following options:
```
{
    message,
    subject,
    email,
    social: Share.Social.EMAIL,
    url: "file:// some path",
    type: 'application/octet-stream'
}
```

2. Observe email app being opened

### Expected behaviour
The email should contain both message and the file included in `url`. However, on iOS, this will just append the file URI to the end of the message. On Android, the attachment is included just fine.

Additionally, if we use `open` instead of `shareSingle` and pick the email client, the file is included just fine on both iOS and Android, but if we do this, we cannot set the email to address.

### Actual behaviour
File attachments should work on both android and iOS

### Environment
- **React Native version**: 0.62.2
- **React Native platform + platform version**: iOS 10+, Android 6+

### react-native-share
**Version**: 3.7.0

### Link to repo (highly encouraged)
-

cristianoccazinsp:
Additionally, `Share.open` should set the default email when `email` is given, just like `shareSingle` (which is by the way incorrectly documented).

IndianNative:
This isssue is now consistent in both android and ios, email doesn't attach the attachment

peterchen2cfa:
any update on this?

PR #847, opened 2020-08-24
Allow shareSingle to share attachments through emails

cristianoccazinsp:
# Overview

`shareSingle` on iOS does not support sharing files like Android. This PR changes the sharing logic to use allow for files to also be shared.  

Note that this changes the way email sharing works with `shareSingle`.  That is, instead of opening the email app with an URL, it uses MFMailComposeViewController (similar to what the `open` call uses) to open the email view right on the app. One may consider this as a breaking change since it uses the app's UI layout tree.  For example, if share is called in a modal and the modal is closed, the share view will also close. However, it is a much smoother experience and matches the email behaviour when the `open` call is used.

# Test Plan

Tested on iPhone 7.

Tested attaching local files (with `url` and `type`)
Tested attaching a base64 file as ````data:image/png;base64,${b64}```` with and without the `filename` property.


cristianoccazinsp:
Should close https://github.com/react-native-community/react-native-share/issues/846

ios/EmailShare.m
```diff
@@ -32,23 +63,109 @@ - (void)shareSingle:(NSDictionary *)options
...
+//            break;
+//        case MFMailComposeResultSent:
+//            NSLog(@"Mail sent");
+//            break;
+//        case MFMailComposeResultFailed:
+//            NSLog(@"Mail sent failure: %@", [error localizedDescription]);
+//            break;
+//        default:
+//            break;
+//    }
```
MateusAndrade, line 163: We can remove this function?

ios/EmailShare.m
```diff
@@ -10,18 +10,49 @@
 
 
 @implementation EmailShare
+
+- (NSString*)getExtensionFromBase64:(NSString*)base64String {
```
MateusAndrade, line 14: I think we already have a function that handles `base64` , maybe we can give a look to avoid duplicated code.

ios/EmailShare.m
cristianoccazinsp: We need the handler so the share/email view can be closed after it's used.  We can certainly remove the commented code though.  I was wondering if there could be any use for it, but then I realized that anything done here wouldn't really be compatible with the rest of the share alternatives nor its Android version.

For example, this could be used to trigger the success event only after the user has closed/sent the email, and we could pass an enum to indicate what did the user do, but this wouldn't be possible on Android...

ios/EmailShare.m
cristianoccazinsp: You are right, there's something similar, but it returns a hardcoded PNG if there are no matches, and it also performs a file save operation from an extra parameter containing data.  I didn't want to break that behaviour or add a bunch of different if conditions to handle each use.  If you can take a look and provide some recommendation it would be good.

ios/EmailShare.m
```diff
@@ -32,23 +63,109 @@ - (void)shareSingle:(NSDictionary *)options
...
+        [mc setToRecipients:@[email]];
+        [mc setSubject:subject];
+        [mc setMessageBody:message isHTML:NO];
+        
         if ([options objectForKey:@"url"] && [options objectForKey:@"url"] != [NSNull null]) {
-            message = [message stringByAppendingString: [@" " stringByAppendingString: [RCTConvert NSString:options[@"url"]]] ];
-        }
-
-        NSString * urlWhats = [NSString stringWithFormat:@"mailto:%@?subject=%@&body=%@", email, subject, message ];
-        NSURL * whatsappURL = [NSURL URLWithString:[urlWhats stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding]];
```
jgcmarins, line 41: can we keep this?

cristianoccazinsp: They are not really equivalent, and it's either one or the other.  What would be the rationale to keep it?  We might need a new flag/setting to enable both options.

cristianoccazinsp: Looking at it again, these lines are in fact redundant and can be achieved with the same `Linking` API that's already part of RN.  So it makes even more sense for the Share API to use an improved version to send emails.

ios/EmailShare.m
cristianoccazinsp: I've removed the comments.

ios/EmailShare.m
cristianoccazinsp: Moved this to an utils class.

cristianoccazinsp:
Any updates?

cristianoccazinsp:
Did I miss any of the required review actions?

cristianoccazinsp:
Any news? It has been a while already.

MateusAndrade:
> Any news? It has been a while already.

Sorry for the delay, I was having some problems focusing on other projects. Btw your PR looks great! We can go on merging this! 

Thanks! 🚀 🚀 🚀 🚀 🚀

MateusAndrade (approved):

t08-apporio:
hi @cristianoccazinsp. I am trying to shareSingle a url using Email. On android it works as expected but on iOS the url is not added to the body. The url is to an external site. Any idea what could be wrong?

cristianoccazinsp:
@t08-apporio can you share some code of what you're doing? Also iOS version, RN version, and react-native-share version.

t08-apporio:
Hi @cristianoccazinsp, here are the specifics

```
const shareEmail = {
 title: "title",
message: "message",
url: "https://example.com",
social: Share.social.EMAIL,
subject: "subject"
}
```
```
Share.shareSingle(shareEmail)
.then(response => handleResponse(response)
.catch(error => alert(error));
```
iOS: 14.0
RN: 0.63.4
react-native-share: 4.1.0

cristianoccazinsp:
I believe you need to use `email` to set the email's TO address when using `shareSingle` instead of `url`.

t08-apporio:
the objective is to share a url via email. Email's "to" address needs to be added by the user after the mail app opens up. The message, subject are filled just fine. Only issue is that the url is not added to the body @cristianoccazinsp

cristianoccazinsp:
In that case, I think you should be adding the URL manually to the email's body? Otherwise, it may be a bug, but it is not related to this PR at all. This PR just added attachments support to `shareSingle`.

t08-apporio:
ah shucks! thanks for your input @cristianoccazinsp. I'll open a new issue if its not already present

cristianoccazinsp:
You should definitely pre-fill the body with the share URL, any reason not to do so? You are sharing with email specifically so it makes sense to build the body yourself. I believe the issue is the same I faced with attachments when I started this PR.

t08-apporio:
@cristianoccazinsp is it possible to pre-fill the mail body? I did not come across it in the documentation. Also I removed the message attribute and tried but that resulted in the email view not opening

cristianoccazinsp:
Yes, setting `message` to some text should prefill the email body.

t08-apporio:
@cristianoccazinsp yeap. I was under the misconception that the message field will be used by other social media platforms. Anyway, thanks a lot for your help. If you don't mind, have you tried sharing on LinkedIn? On iOS it doesn't do anything, but on android it works

cristianoccazinsp:
@t08-apporio sorry, I haven't really tested LinkedIn, but I wouldn't be surprised! The iOS side of this library is quite buggy, but should be working fine when using the generic share (when the user selects the platform).

PR #856, opened 2020-09-03
Fixes for some bugs with Instagram Stories on Android

### LINKED ISSUE #852: Algorithm to break up base64 content is too slow
I have a fix for this. Just opening the issue to remind me to post a PR.

### Steps to reproduce
1. Use a large file
2. Set the attachment to use base64 encoded
3. Share.

### Expected behaviour
File should send

### Actual behaviour
Stack overflow

### Environment
- **React Native version**: 
- **React Native platform + platform version**: iOS 9.0, Android 5.0, etc

### react-native-share
**Version**: 3.7.0

https://github.com/

### LINKED ISSUE #853: `super.open` is called twice when doing `shareSingle`
Again, I have a fix for this. Just an issue to remind me to PR it.

Android-specific issue.

When you run `shareSingle`, it calls `super.open` once, with the correct data. Then, it calls it again further down.

It doesn't break anything, but the processing all runs twice for no reason.

### LINKED ISSUE #854: [android] FLAG_GRANT_READ_URI_PERMISSION is overwritten before the intent is sent in singleShareIntent
I have a fix for this one. Another issue to remind me to do a PR.

This happens only when `forceDialog` is `false` (or unset)

When you share with instagram, the `FLAG_GRANT_READ_URI_PERMISSION` is set for the background image. However, `this.getIntent().setFlags()` is called again before shipping, which overrides the existing flags. 

This can be fixed pretty easily by using `addFlags` instead of `setFlags` before calling `startActivity`

### LINKED ISSUE #855: [Android] Impossible to share an Instagram Story sticker from URI if `forceDialog` is false
On android, when using `shareSingle` on Instagram Stories, you need to `activity.grantUriPermission` to the URI sent through for the sticker asset. This only happens when `forceDialog` is set to `true`.

A fix for this is to call the `activity.grantUriPermission` when setting the URI, instead of waiting until right before sending the intent.

I have a fix for this. Issue is just to remind me to do this.

deecewan:
# Overview

There were a couple of shortcomings when trying to share Instagram Stories on Android. `base64` would _not_ work on a file of any decent size. The flags on the intents were not set correctly, so Instagram was not able to open the image after it was sent. Stickers could only be sent if you were using `forceDialog`.

I've just upstreamed the changes I made at work to get Instagram Stories to work in our app.

Fixes #852 
Fixes #853 
Fixes #854 
Fixes #855

# Test Plan

There are no tests for Android currently. I have patched these changes into the app at work, and they all work as expected (for us), so I'm upstreaming them

### DIFF
--- a/android/src/main/java/cl/json/ShareFile.java
+++ b/android/src/main/java/cl/json/ShareFile.java
@@ -16,6 +16,8 @@ import java.io.IOException;
  */
 public class ShareFile {
 
+    public static final int BASE_64_DATA_LENGTH = 5; // `data:`
+    public static final int BASE_64_DATA_OFFSET = 16; // `;base64,` + 8
     private final ReactApplicationContext reactContext;
     private String url;
     private Uri uri;
@@ -54,14 +56,25 @@ public class ShareFile {
     public boolean isFile() {
         return this.isBase64File() || this.isLocalFile();
     }
+
     private boolean isBase64File() {
         String scheme = uri.getScheme();
         if((scheme != null) && uri.getScheme().equals("data")) {
-            this.type = this.uri.getSchemeSpecificPart().substring(0, this.uri.getSchemeSpecificPart().indexOf(";"));
+            StringBuilder type = new StringBuilder();
+            char[] parts = this.uri.toString().substring(BASE_64_DATA_LENGTH).toCharArray();
+            for (char part : parts) {
+                if (part == ';') {
+                    break;
+                }
+                type.append(part);
+            }
+
+            this.type = type.toString();
             return true;
         }
         return false;
     }
+
     private boolean isLocalFile() {
         String scheme = uri.getScheme();
         if((scheme != null) && (uri.getScheme().equals("content") || uri.getScheme().equals("file"))) {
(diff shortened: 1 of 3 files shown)


MateusAndrade (approved):
WOW! Thanks a lot, @deecewan. 🚀

android/src/main/java/cl/json/ShareFile.java
```diff
@@ -16,6 +16,8 @@
  */
 public class ShareFile {
 
+    public static final int BASE_64_DATA_LENGTH = 5; // `data:`
+    public static final int BASE_64_DATA_OFFSET = 16; // `;base64,` + 8
```
sterlingwes, line 20: curious how you tested this? I'm finding that unless this value is `8`, it leads to a corrupted file. (I'm working with a base64-encoded PDF for what it's worth)

deecewan: ah. i might have cooked it! you're right. i misread the code that i was translation (from [here](https://github.com/react-native-community/react-native-share/pull/856/files#diff-2fa2788e1b71bd42f81abefbb76a001bL109)). I don't know how this was working in my local implementation 🤔.

Issue #839, opened 2020-08-04
Sharing a Base-64 encoded PDF on Android is broken on targetSdk 29


WaterCityFlame:
### Steps to reproduce
1. Change your android/build.gradle so the following lines use version 29: 
```
        buildToolsVersion = "29.0.0"
        compileSdkVersion = 29
        targetSdkVersion = 29
```
2. Add a button or something to share a PDF file that was base-64 encoded
```javascript
/*options like */
const options = {
    url: "date:application/pdf;base64, /*something,something,file */"
    type: "application/pdf"
}

/* on press function like: */
onPress = () => {
         Share.open(options)
              .then((res) => {
                  console.log(res);
              })
              .catch((err) => {
                  err && console.log(err);
              })
}
```
3. Build your app
4. Press the button
5. Allow file system access when Android asks
6. Select an app to share it with (I use GMail usually)

### Expected behaviour
I expect the selected app to open with the PDF file. E.g. GMail should have it already attached to a new message

### Actual behaviour
The selected app opens, but there is no PDF. E.g. GMail opens to write a new message, but there is no attachment

### Environment
- **React Native version**: "0.63.2" - in another project using "0.61.5" I also reproduced this 
- **React Native platform + platform version**: Android 10.0

### react-native-share
**Version**: 3.7.0 - the one you get running `npm i react-native-share --save`

### Link to repo
https://github.com/WaterCityFlame/PdfSharingAndroid
The branch target-sdk-28 shows that it **does** work when your targetSdk is 28. 

### Other Questions You May Have: 
**Does this impact all base-64 encoded files or just PDFs?"** I don't actually know, will try to reproduce that tonight. 
**Does using the "filename" option instead of the "url" option fix it?"** Nope. I don't have a work-around at present
**If you include other options, do they work?"** message does, don't know about others.

adrianm86:
Sharing png and jpeg images with base64 in react-native 0.63.2 also does not work.
I've updated to this new version of react-native and it doesn't work anymore.

WaterCityFlame:
@adrianm86 what's in your android/build.gradle file at the moment? My working theory was that the cause was the targetSdk, not the react-native version directly (although RN may update that)

adrianm86:
Yes, I also have targetSdk 29.
Since version 0.63 of react-native the default Android targetSdk is version 29.

mutablestudio:
[...]

PR #870, opened 2020-09-17
fix(android): avoid corrupting base64 files w/ Share.open

sterlingwes:
# Overview
Follow-on to https://github.com/react-native-community/react-native-share/pull/856, I noticed that sharing a base64'd PDF would lead to a corrupted file since we're dropping the first 8 characters from the data part of the URI. Hoping @deecewan can confirm whether this is compatible with this work / whether 16 was in fact intended.

Might be related to https://github.com/react-native-community/react-native-share/issues/839

# Test Plan

Added a base64 dummy PDF to the sample app and confirmed it opens in Drive in my emulator

pdftest


deecewan:
Hey 👋 this should be fine. IIRC i only did that to make it match the existing behaviour the code already had. If this works, I hav e no objections to it 😄

MateusAndrade (approved):
Nice! Thanks 🙉 !

PR #872, opened 2020-09-18
chore(deps): commit lockfile

sterlingwes:
# Overview
I noticed the yarn lockfile changed unexpectedly on my other branch so I'm committing it separately. I think someone may have just forgotten it

But just in case:

```
▶ node -v
v12.16.3

▶ yarn --version
1.22.4
```

# Test Plan
None


sterlingwes:
Seems like this diff landed in 95b86e712552b081b3ffb270c7c051b282b269e5

Issue #879, opened 2020-09-29
App crash (cannot read property Version of undefined)


alemarra89:
### Steps to reproduce
1. Update to react-native-share v4
2. Clean project (node_modules, pod install, npm start -- --reset-cache, ecc)
3. Build and launch (simulator or real device)

### Expected behaviour
It works

### Actual behaviour
App crashed

### Environment
- **React Native version**: 0.60.6
- **React Native platform + platform version**: iOS 14.0 on simulator, Android 11 on Google Pixel 4, iOS 13.5.1 on iPhone X

### react-native-share
**Version**: "master" (version 4)

With debugger, I discovered the problem is in these lines:

> const ANDROID_KIT_KAT_SDK_VERSION = 19;
> const androidPermissionRequestRequired = Platform.constants.Version < ANDROID_KIT_KAT_SDK_VERSION;

Cannot read property Version of undefined

I hope there is everything to reproduce the error.

sterlingwes:
Ah jeez... `Platform.constants` was added in RN 61...

MateusAndrade:
> Ah jeez... `Platform.constants` was added in RN 61...

Maybe we can update the migration guide, or force a `||`. What do you think?

alemarra89:
I can confirm that app works correctly now, with react-native-share@4.0.1

PR #880, opened 2020-09-29
fix: use documented Platform.version for backwards compat

sterlingwes:
# Overview

The `Platform.constants` added in #871 is not available in RN versions before 0.61 as reported in #879. `Platform.Version` is still there, and it's documented so I've switched to that instead.

# Test Plan

Going to ask @alemarra89 if they can install this branch and test that it resolve the issue.

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,7 +22,7 @@ import Sheet from './components/Sheet';
 import Button from './components/Button';
 
 const ANDROID_KIT_KAT_SDK_VERSION = 19;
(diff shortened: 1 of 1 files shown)


MateusAndrade (approved):

sterlingwes:
Btw this'll affect both platforms @MateusAndrade I think

MateusAndrade:
> Btw this'll affect both platforms @MateusAndrade I think

Sure, since it's in JS code even `iOS` will get this error. :/

PR #892, opened 2020-10-18
fix: homepage URL in package.json

mars-lan:
# Overview

# Test Plan


mikehardy (approved):
The great react-native-community dispersion weeeee

PR #938, opened 2020-12-23
fix typo in local variable name: sticketURL -> stickerURL

set0gut1:
# Overview

I found a typo in the local variable name, so I fixed it.

# Test Plan


MateusAndrade (approved):
Thanks!

PR #939, opened 2020-12-23
Proposal: Removal of WRITE_EXTERNAL_STORAGE permission

ryanbourneuk:
# Overview

Fixes proposal https://github.com/react-native-share/react-native-share/issues/936.

I've removed the `WRITE_EXTERNAL_STORAGE` permission from the module's AndroidManifest.xml file (and added this to the example project), as it should be the responsibility of the implementing project to add this permission if base64 file sharing is a requirement. I have also updated the documentation to mention this change.

As detailed on the issue, I see this as a breaking change (to prevent the possibility of implementing projects picking a new version with this included, and apps breaking unexpectedly), so I have included additional migration documentation too.

Please let me know if there's any changes required!

# Test Plan

Upon opening the example project, it should be possible to share base64 files, in the same manner as before this change was introduced.

### DIFF
--- a/website/docs/migrate-v4-to-v5.mdx
+++ b/website/docs/migrate-v4-to-v5.mdx
@@ -11,6 +11,8 @@ No breaking changes :relieved:
 
 [v5 introduces a change](https://github.com/react-native-community/react-native-share/issues/936) to remove the `WRITE_EXTERNAL_STORAGE` permission from being automatically added to implementing apps.
 
+This is motivated by the Android 11 system enforcing new restrictions on storage, and mandating the use of scoped storage. If this impacts you, you should read more here to make sure your app will still work on Android 11: [Scoped Storage restrictions](https://developer.android.com/about/versions/11/privacy/storage)
+
 ### Removal of WRITE_EXTERNAL_STORAGE permission request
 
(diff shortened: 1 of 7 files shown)


MateusAndrade (dismissed):
LGMT! Do you guys have any considerations: @jgcmarins , @mikehardy ?

website/docs/migrate-v4-to-v5.mdx
```diff
@@ -0,0 +1,25 @@
...
+---
+
+## iOS
+
+No breaking changes :relieved:
+
+## Android
+
+[v5 introduces a change](https://github.com/react-native-community/react-native-share/issues/936) to remove the `WRITE_EXTERNAL_STORAGE` permission from being automatically added to implementing apps.
+
```
mikehardy, line 13:
```suggestion

This is motivated by the Android 11 system enforcing new restrictions on storage, and mandating the use of scoped storage. If this impacts you, you should read more here to make sure your app will still work on Android 11: [Scoped Storage restrictions](https://developer.android.com/about/versions/11/privacy/storage)

```

mikehardy (changes requested):
that permission basically must be removed for targetSdk 30, so if the package can function without it, it should not include the permission

So I really like this change

I am marking it for request changes only because this is such an important thing I wanted to add the reason why along with a link to the underlying doc

@ryanbourneuk -> excellent otherwise

website/docs/migrate-v4-to-v5.mdx
ryanbourneuk: @mikehardy I've added the updated docs in a11464e08ee4df5c401a019f40a3d6102f8c94aa 😄

mikehardy (approved):
Ace

MateusAndrade (approved):

PR #941, opened 2020-12-23
LSApplicationQueriesSchemes documentation

farad4y:
This is mentioned on a different page, but this information is not included on Share.shareSingle method docs. It took me quite a while to understand the problem, I think it would be helpful to put this information on this page.


MateusAndrade (approved):
Thanks!

PR #964, opened 2021-02-08
Make appId optional in Options type

grit96:
The `appId` option isn't mentioned in the [docs](https://react-native-share.github.io/react-native-share/docs/share-open/#supported-options) but is marked as mandatory in the Options flow type which breaks flow type checking.


MateusAndrade (approved):
Thanks!

PR #968, opened 2021-02-20
Android: Fix permission denial error

roman-r-m:
Fixes the error from https://github.com/react-native-share/react-native-share/issues/613 in Android 11.


MateusAndrade (approved):
Thanks, @roman-r-m! 🚀

Issue #857, opened 2020-09-03
Instagram Stories - No Success Callback (iOS only)


AechDub:
### Steps to reproduce
1. Attempt to share to instagram stories
2. Return to app
3. Notice no success callback is triggered.

### Expected behaviour
As other share behaviours, a success callback should trigger

### Actual behaviour
We have a thread waiting for return that never completes.

### Environment
- **React Native version**: 0.62.2
- **React Native platform + platform version**: iOS 13.6.1

### react-native-share
**Version**: 3.7.0

### Link to repo (highly encouraged)
n/a

AechDub:
Stalebot shouldn’t drop tickets like this. We have people who have made actions on this to follow. :)

MateusAndrade:
> Stalebot shouldn’t drop tickets like this. We have people who have made actions on this to follow. :)

Hi @AechDub, did you find what was causing that at your project? I think it's related to other issues with Instagram Shar. If you found anything that may help us debugging that It would be great to add this to the docs(or maybe finding a fix) 😃

ksairi:
@MateusAndrade @AechDub 
Hi all! I'm having an strange issue with this lib, I take a any example that should work like this one:
```
try {
          const shareOptions = {
            method: Share.FacebookStories.SHARE_BACKGROUND_AND_STICKER_IMAGE, // iOS only
            backgroundVideo: 'URI_TO_MP4', // Android only (uri to a local file)
            backgroundImage: 'http://urlto.png', // url or an base64 string
            stickerImage: 'data:image/png;base64,<imageInBase64>', // or you can use "data:" url
            backgroundBottomColor: '#fefefe',
            backgroundTopColor: '#906df4',
            attributionURL: 'http://deep-link-to-app', // in beta
            appId: '219376304', // facebook appId
            social: Share.Social.FACEBOOK_STORIES,
          };

          await Share.shareSingle(shareOptions);
        } catch (e) {
          console.log(e);
        }
```
and I don't get any response at all, no errors on catch and if I do 
```
       const response = Share.shareSingle(shareOptions);
       console.log(response)
```
`{}` gets printed
This is happening with `Share.Social.*`, This is in iOS 14.1 and latest lib version
any ideas?

PR #969, opened 2021-02-22
fix: resolve promise on ios every time

### LINKED ISSUE #865: Promise "open" is never resolved when canceling printing
### Steps to reproduce
1. Open a image in Base64, like this:
```js
                 const openResult = await Share.open({
                        failOnCancel: true,
                        title: 'Some title',
                        url: 'data:image/jpg;base64,foobarbazz',
                        filename: 'file-name.jpg',
                    });
```

2. Select "Print" option.

image

3. Then click on "Cancel" option.

image

### Expected behaviour

The promise should be rejected (failOnCancel=true) or resolved (failOnCancel=false).

### Actual behaviour

The promise never end.

### Environment

- **React Native version**: 0.61.5
- **React Native platform + platform version**: iOS 13.7

### react-native-share

**Version**: 3.8.0

I don't think it's necessary.

krystofcelba:
# Overview
The `Share.open` promise doesn't resolve in some situations on iOS. For example when you cancel printing. It's well described in #865. This PR fixes that.

Fix #865 #857

# Test Plan

Please check #865, there are steps for reproduce and it should work as expected with this fix.

### DIFF
--- a/ios/RNShare.m
+++ b/ios/RNShare.m
@@ -266,7 +266,7 @@ RCT_EXPORT_METHOD(open:(NSDictionary *)options
         if (activityError) {
             [controller  dismissViewControllerAnimated:true completion:nil];
             failureCallback(activityError);
-        } else if (completed || activityType == nil) {
+        } else {
             successCallback(@[@(completed), RCTNullIfNil(activityType)]);
         }
     };


ios/RNShare.m
```diff
@@ -266,7 +266,7 @@ - (NSDictionary *)constantsToExport
         if (activityError) {
             [controller  dismissViewControllerAnimated:true completion:nil];
             failureCallback(activityError);
-        } else if (completed || activityType == nil) {
```
krystofcelba, line 269: The condition here is unnecessary as the completion is checked [https://github.com/krystofcelba/react-native-share/blob/80fc9967efb4162e19950eb0f209bc2e14a49231/index.js#L268](here) in JS code and in situations like canceled print it causes that the JS callback isn't called at all.

MateusAndrade (approved):
Nice!!! Thanks, @krystofcelba! 🚀 🚀 🚀

cristianoccazinsp:
This introduced the following bug: https://github.com/react-native-share/react-native-share/issues/972

I'm already testing a "fix", but it would be great to get @krystofcelba input.

cristianoccazinsp:
Proposed update: 

```
__weak UIActivityViewController* weakShareController = shareController;
    shareController.completionWithItemsHandler = ^(NSString *activityType, BOOL completed, __unused NSArray *returnedItems, NSError *activityError) {
        
        // always dismiss since this may be called from cancelled shares
        // but the share menu would remain open, and our callback would fire again on close
        [controller dismissViewControllerAnimated:true completion:nil];
        
        if (activityError) {
            failureCallback(activityError);
        } else {
            successCallback(@[@(completed), RCTNullIfNil(activityType)]);
        }
        
        // clear the completion handler to prevent cycles
        if(weakShareController){
            weakShareController.completionWithItemsHandler = nil;
        }
    };
```

This has the side effect that the share UI is always closed if the selected app was "cancelled".  However, it doesn't seem possible to detect if the share UI was closed automatically by the system or not

PR #984, opened 2021-03-20
Android: fix sharing a single file

roman-r-m:
This is a follow-up to my previous PR https://github.com/react-native-share/react-native-share/pull/968 - looks like I have introduced a bug there.
According to the [official dev guide](https://developer.android.com/training/sharing/send.html#send-binary-content) we only need to set the type, and the URI goes to `EXTRA_STREAM`.

Without this change if I share a file to, say Gmail, I get a new email with the subject containing the URI which is clearly wrong. After the fix, as expected, I get a new email with the file attached.

Fixes https://github.com/react-native-share/react-native-share/issues/976

### DIFF
--- a/android/src/main/java/cl/json/social/ShareIntent.java
+++ b/android/src/main/java/cl/json/social/ShareIntent.java
@@ -252,7 +252,7 @@ public abstract class ShareIntent {
             this.fileShare = getFileShare(options);
             if (this.fileShare.isFile()) {
                 Uri uriFile = this.fileShare.getURI();
-                this.getIntent().setDataAndType(uriFile, this.fileShare.getType());
+                this.getIntent().setType(this.fileShare.getType());
                 this.getIntent().putExtra(Intent.EXTRA_STREAM, uriFile);
                 this.getIntent().addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);
                 if (!TextUtils.isEmpty(message)) {


MateusAndrade (approved):

MateusAndrade:
Thanks!

yfunk:
@roman-r-m @MateusAndrade It seems like this PR introduced a regression of the fix in the previous PR #968.

I was getting the `SecurityException` on `v6.3.0` and found the previous PR during troubleshooting. I then downgraded the library `v5.1.3` (where the first previous PR was added) and the error was gone. After upgrading to `v5.1.5` (next change to `android/src/main/java/cl/json/social/ShareIntent.java`) which added this PR the error was getting thrown again.
```
2021-07-19 16:09:51.019 7918-7942/com.myapp E/DatabaseUtils: Writing exception to parcel
    java.lang.SecurityException: Permission Denial: reading cl.json.RNShareFileProvider uri content://com.myapp.rnshare.fileprovider/rnshare2/export/test.zip from pid=314, uid=1000 requires the provider be exported, or grantUriPermission()
        at android.content.ContentProvider.enforceReadPermissionInner(ContentProvider.java:729)
        at android.content.ContentProvider$Transport.enforceReadPermission(ContentProvider.java:602)
        at android.content.ContentProvider$Transport.query(ContentProvider.java:231)
        at android.content.ContentProviderNative.onTransact(ContentProviderNative.java:104)
        at android.os.Binder.execTransactInternal(Binder.java:1021)
        at android.os.Binder.execTransact(Binder.java:994)
```
Although I didn't notice any issues related to sharing with a Base64 image or ~250MB zip archive.

**Environment:**
```
react-native: 0.64.2
react-native-share: 5.1.5 - 6.3.0
```
```
minSdkVersion = 21
compileSdkVersion = 30
targetSdkVersion = 30
```
Tested on Emulator (API 29) and Samsung Galaxy M20 (API 29)

PR #994, opened 2021-04-07
Added TS type declarations from @types/react-native-share

taeh98:
This copies the type declaration files from @types/react-native-share into the repo and links them to package.json.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,134 @@
+// Type definitions for react-native-share 3.3
+// Project: https://github.com/react-native-community/react-native-share#readme
+// Definitions by: Mark Nelissen <https://github.com/marknelissen>
+//                 pera <https://github.com/santiagofm>
+//                 MateusAndrade <https://github.com/MateusAndrade>
+//                 Jesse Katsumata <https://github.com/Naturalclar>
+//                 Fabian Lee <https://github.com/fabianlee1211>
+//                 Evan Cloutier <https://github.com/evancloutier>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.1
+import { ReactNode, Component } from 'react';
+import { ViewProps, StyleProp } from 'react-native';
+import Overlay from './Overlay.d';
+import Button from './Button.d';
+import Sheet from './Sheet.d';
+
+declare namespace Share {
+    enum Social {
+        FACEBOOK = 'facebook',
+        PAGESMANAGER = 'pagesmanager',
+        TWITTER = 'twitter',
+        WHATSAPP = 'whatsapp',
+        INSTAGRAM = 'instagram',
+        INSTAGRAM_STORIES = 'instagram-stories',
+        GOOGLEPLUS = 'googleplus',
+        EMAIL = 'email',
+        PINTEREST = 'pinterest',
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -10,13 +10,20 @@
...
     "ios",
     "windows",
     "RNShare.podspec"
   ],
+	"peerDependencies": {
+		"@types/react": "*",
+		"@types/react-native": "*",
+		"react": "*",
+		"react-native": "*"
+	},
```
MateusAndrade, line 26: Why do we need these declarations? This would work without it?

index.d.ts
```diff
@@ -0,0 +1,134 @@
+// Type definitions for react-native-share 3.3
+// Project: https://github.com/react-native-community/react-native-share#readme
```
MateusAndrade, line 2: Maybe we could remove this and set a comment regarding the migration? What do you think?

MateusAndrade (changes requested):

MateusAndrade:
@taeh98 can you fill in a description on the PR?

package.json
taeh98: I copied and pasted them from the commit in another React native project that added TS types. Now removed.

index.d.ts
taeh98: New comment now written.

taeh98:
> @taeh98 can you fill in a description on the PR?

Done.

MateusAndrade:
> > @taeh98 can you fill in a description on the PR?
> 
> Done.

Nice! Btw, we need to run a `yarn` to update the `yarn.lock`, otherwise, we will end with an error on the lock.

taeh98:
> > > @taeh98 can you fill in a description on the PR?
> > 
> > 
> > Done.
> 
> Nice! Btw, we need to run a `yarn` to update the `yarn.lock`, otherwise, we will end with an error on the lock.

Done in https://github.com/react-native-share/react-native-share/pull/994/commits/3a96139b1a264427ca7e4fda5286f7bb814054ee

MateusAndrade (approved):
Thanks!!!

DanielMarkiel:
Hi @MateusAndrade,
I just upgraded to v5.2.0 and I can't find type declarations in the package (no `index.d.ts` file in `node_modules/react-native-share`).
Are you sure they are exported correctly?

Ps. I tried to remove `node_modules` and install them again (using yarn v1.22.10) - same effect.

image

MateusAndrade:
> Hi @MateusAndrade,
> I just upgraded to v5.2.0 and I can't find type declarations in the package (no `index.d.ts` file in `node_modules/react-native-share`).
> Are you sure they are exported correctly?
> 
> Ps. I tried to remove `node_modules` and install them again (using yarn v1.22.10) - same effect.
> 
> image

I think we need to publish then to npm with the other packages, we will need to change this: https://github.com/react-native-share/react-native-share/blob/f5521caa264455e2f777c4da20f5d158bcf6f0bd/package.json#L14-L20 adding the `*.ts` files

mikehardy:
Also a types field I think? https://github.com/react-native-device-info/react-native-device-info/blob/7209edc41d8323796324f2ecc3459d50d4978308/package.json#L6

MateusAndrade:
Fixing the missing publish config at #997

PR #996, opened 2021-04-09
Added a quick summary of the open() and shareSingle() methods to docs

taeh98:
As someone who is new to the library, I struggled to understand what the open() and shareSingle() methods do and the difference between them. I think adding these short descriptions to the docs may help others in the same position.


MateusAndrade (approved):
Thanks! 🚀

PR #1001, opened 2021-04-16
feat(typescript): migrate project to typescript

DanielMarkiel:
# Overview
Introduces typescript to project

See #805 and https://github.com/react-native-share/react-native-share/pull/818

@tcodes0 Thanks for your input!

# Test Plan
Run `yarn validate`

# TODO

- [x]  deps and config
- [x]  components
- [x]  eslint-typescript configuration, should fix ci
- [x]  figure out how to build ts for npm
- [x]  type index.js
- [x]  remove flow
- [x]  read bridge and double check types


ios/RNShare.m
```diff
@@ -104,12 +104,12 @@ - (NSDictionary *)constantsToExport
 {
   return @{
     @"FACEBOOK": @"facebook",
-    @"FACEBOOK_STORIES": @"facebookstories",
+    @"FACEBOOKSTORIES": @"facebookstories",
```
EstebanFuentealba, line 107: Why did this change?

DanielMarkiel: To standardize naming between iOS and Android. 
In android we've got `facebookstories` (which is later uppercased to `FACEBOOKSTORIES`).
Original implementation -> `NativeModules.RNShare.FACEBOOK_STORIES` was returning undefined on Android (because the key name there is `FACEBOOKSTORIES`).

prettier.config.js
```diff
@@ -1,5 +0,0 @@
-module.exports = {
```
EstebanFuentealba, line 1: I don't know if deleting prettier.config.js is a good idea. Because in future changes the code will be indented with the configuration of each code editor?

MateusAndrade:
What about keeping the flow types @DanielMarkiel? How do you think we would be able to achieve this or do you think it's worth it to keep these types?

prettier.config.js
DanielMarkiel: Configuration was moved to `.eslintrc.js` file.
I've got a prettier plugin in my editor and it was recognizing the config correctly. After this change `yarn lint` will let you know if something is wrong even if you don't have a plugin. Also CI (`yarn validate`) will be double-checking that everything is ok.

DanielMarkiel:
@MateusAndrade I agree with the @mikehardy 's opinion presented here -> https://github.com/react-native-share/react-native-share/issues/805#issuecomment-645499711.
If there's a need for flow types the best way would be to put them into flow-typed repo.
It should be fairly easy now when the package is typed in typescript (first tool from the google search -> https://github.com/joarwilk/flowgen)

EstebanFuentealba:
Good job, maybe it would be good to add an extension to the definition of  NativeModules like this

```js
declare module 'react-native' {
  export interface RNShare {
    open: (config: ShareOptions, /* TODO: ... */) => Promise<void>;
    /* TODO */
  }
  interface NativeModulesStatic {
    RNShare: RNShare;
  }
}

```

DanielMarkiel:
@EstebanFuentealba Great suggestion! :) Done

MateusAndrade:
Nice!!! Do you guys think we should bump this as a major version?

cc: @EstebanFuentealba , @DanielMarkiel

DanielMarkiel:
> Nice!!! Do you guys think we should bump this as a major version?
> 
> cc: @EstebanFuentealba , @DanielMarkiel

Yep, in my opinion it should be a major version bump

MateusAndrade (approved):

MateusAndrade:
Do you think we can go on with this @EstebanFuentealba?

src/types.ts
```diff
@@ -81,12 +68,12 @@ interface BaseSocialStoriesShareSingleOptions extends Omit<BaseShareSingleOption
...
 export interface InstagramStoriesShareSingleOptions extends BaseSocialStoriesShareSingleOptions {
   social: Social.InstagramStories;
-  method: InstagramStories;
+  method: ShareAsset;
 }
 
 export interface FacebookStoriesShareSingleOptions extends BaseSocialStoriesShareSingleOptions {
   social: Social.FacebookStories;
-  method: FacebookStories;
+  method: Exclude<ShareAsset, ShareAsset.BackgroundVideo>;
```
MateusAndrade, line 76: Nice!

MateusAndrade:
@DanielMarkiel did you test this with the example app?

DanielMarkiel:
@MateusAndrade Yes, I did. Android looked fine but I couldn't test the iOS version properly since I don't have an iPhone.
I would appreciate it if someone did a retest and double-checked that the package is still working as expected.

MateusAndrade:
> @MateusAndrade Yes, I did. Android looked fine but I couldn't test the iOS version properly since I don't have an iPhone.
> I would appreciate it if someone did a retest and double-checked that the package is still working as expected.

Tested here with iOS and it's working! 🚀 

I'm going to merge this and publish a major version. Thanks a lot, @EstebanFuentealba, @DanielMarkiel!!

PR #1002, opened 2021-04-19
Adding mavenCentral() as jcenter() is shutting

maheshwarimrinal:
Adding mavenCentral() as jcenter() is shutting down


mikehardy (approved):
Just did similar in react-native-device-info, and I have this in progress in react-native-firebase, works fine for libraries already (like this PR) example app still needs it though as react-native is not ready yet

Issue #603, opened 2019-10-08
WhatsApp: No way to share a local image


Psiiirus:
### Steps to reproduce
1. pass an file:// , non-file:// or base64 into the url or urls into sharing-options
2. try to open sharing
3. select WhatsApp
4. nothing will be appended

### Expected behaviour
It should append the file(s) like it it does an every other channel like instagram, MMS, AirDrop , Email,...

### Actual behaviour
Share.openSingle-function:
- If i add an base64 format it just adds the whole string to the message.
- If i add a local file:// link created by RNBlob.fetch  it just adds the local uri to the message. 
- if i add a *non*-file:// link created by RNBlob.fetch it just adds the local uri to the message. 

Share.open function: 
- If i add an base64 format it just adds the whole string to the message.
- If i add a local file:// link created by RNBlob.fetch  it just does nothing.
- if i add a *non*-file:// link created by RNBlob.fetch it just does nothing.

IMPORTANT: if you remove the file extension and just pass the temp-filename into Share.open it will send this file but will not preview it because of the lack of the image type.

### Environment
- RN60
- iOS13

### react-native-share
2.0

### Link to repo (highly encouraged)
https://github.com/

jgcmarins:
sharing base64 should be just like this:
```tsx
const url = `data:image/png;base64,${base64FileString}`

const shareOptions = {
  title: 'Share example',
  message: 'Awesome Message'
  url: url,
  failOnCancel: false,
};

try {
  await Share.open(shareOptions);
} catch (err) {
  // do something
}
```

Psiiirus:
> sharing base64 should be just like this:
> 
> ```tsx
> const url = `data:image/png;base64,${base64FileString}`
> 
> const shareOptions = {
>   title: 'Share example',
>   message: 'Awesome Message'
>   url: url,
>   failOnCancel: false,
> };
> 
> try {
>   await Share.open(shareOptions);
> } catch (err) {
>   // do something
> }
> ```

thanks for your replay, i think i already tried it like this and it didnt work but i will do it again.

Do you have an app ready and running which can share a local image through whatsApp ?

Psiiirus:
@jgcmarins Unf. it didnt work....works with MMS , iMessage, Insta , etc but not with WhatsApp

jgcmarins:
mind sharing a reproducible repo?

Psiiirus:
> mind sharing a reproducible repo?

[...]

PR #1006, opened 2021-04-20
fix(whatsapp): fix iOS Whatsapp images share

alessioemireni:
# Overview

This PR fix bug [#603](https://github.com/react-native-share/react-native-share/issues/603). With this implementation the _Share.shareSingle_ method will share an Image with Whatsapp.

As described in iOS [Whatsapp Documentation](https://faq.whatsapp.com/iphone/how-to-link-to-whatsapp-from-a-different-app/?lang=en) is not possible to share an image with Custom URL and text with image at the same time with Document Interaction API.

# Test Plan
1. Install Whatsapp on iOS device and verify the account.
2. Open your app
3. Invoke the _Share.shareSingle_ method with these options
```
{
  title: "title",
  subject: "subject",
  url: data:image/jpeg;base64,${base64},
  social: Share.Social.WHATSAPP,
  message: "message"
}
```
4. iOS Document Interaction will open;
5. Choose Whatsapp;
6. A preview of image will open on whatsapp;
7. Click done to share the image to a contact;

### DIFF
--- a/ios/WhatsAppShare.m
+++ b/ios/WhatsAppShare.m
@@ -14,14 +14,14 @@ RCT_EXPORT_MODULE();
 - (void)shareSingle:(NSDictionary *)options
     failureCallback:(RCTResponseErrorBlock)failureCallback
     successCallback:(RCTResponseSenderBlock)successCallback {
-
+    
     NSLog(@"Try open view");
-
+    
     if ([options objectForKey:@"message"] && [options objectForKey:@"message"] != [NSNull null]) {
         NSString *text = [RCTConvert NSString:options[@"message"]];
         text = [text stringByAppendingString: [@" " stringByAppendingString: options[@"url"]] ];
         NSString *whatsAppNumber = [RCTConvert NSString:options[@"whatsAppNumber"]];
-
+        
         if ([[UIApplication sharedApplication] canOpenURL: [NSURL URLWithString:@"whatsapp://app"]]) {
             NSLog(@"WhatsApp installed");
         } else {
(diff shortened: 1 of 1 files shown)


MateusAndrade (approved):

MateusAndrade:
Nice!!! Btw, did we test this on iOS? Unfortunately, I can't test it here. 😢

rumata2:
Guys, thanks for this! It does not solve the Share.open problem though - and in most cases we can't just share directly to Whatsapp, we need to let users choose. Any ideas?

alessioemireni:
> Guys, thanks for this! It does not solve the Share.open problem though - and in most cases we can't just share directly to Whatsapp, we need to let users choose. Any ideas?



> Guys, thanks for this! It does not solve the Share.open problem though - and in most cases we can't just share directly to Whatsapp, we need to let users choose. Any ideas?

This PR is just for `Share.singleShare` method and not `Share.open`. In any case I have to fix an issue with PNG images so I put the PR in draft.

alessioemireni:
The PR now is OK.

MateusAndrade (approved):

samarthjain10:
In this PR Share.shareSingle(shareOptions) is not behaving as expected. Its showing options to chose 
const shareOptions = {
        title: 'some title',
        message: 'some message,
        url: 'base64image',
        social: Share.Social.WHATSAPP,
      }

akdj:
Could we skip the step 4 "iOS Document Interaction will open"  I don't think it is the goal of singleShare with option Whatsapp.  it should generate the .wai file and open whatsapp image preview directly.

PR #1030, opened 2021-05-28
remove build tools version

dulmandakh:
# Overview

Setting buildToolsVersion is no longer recommended, it'll default to Android Gradle Plugin defaults. It makes DX worse by requiring multiple versions of build tools by various dependencies require various versions.

# Test Plan

Everything works as before, no visible changes.


mikehardy (approved):
Agree

PR #1031, opened 2021-06-02
TS: make message optional in Share.open

sseppola:
# Overview

This makes the `message` property on `ShareOptions` optional.

In my experience this property was never necessary, and actually introduced a bug when I followed the TS error. 

Related issues:
- https://github.com/react-native-share/react-native-share/issues/760
- https://github.com/react-native-share/react-native-share/issues/831
- https://github.com/react-native-share/react-native-share/issues/474

# Test Plan

I've omitted the `message` property in my own app (and ignoring the TS error) and it worked fine. Best way to test it is to omit the message property when using `Share.open()`


MateusAndrade (approved):
Nice!!! Thanks. 🚀

PR #1049, opened 2021-07-08
Fix iOS Email URL sharing bug

andyrobert3:
# Overview
Adding URL parameter in `shareSingle` method for `social` type `EMAIL`
https://github.com/react-native-share/react-native-share/issues/932

# Test Plan
Tested on iPhone before and after change

https://www.loom.com/share/d41d13e935774e4190fd44ab731d544d


MateusAndrade (approved):
Thanks! 🚀

PR #1058, opened 2021-07-21
added single share feature for telegram

harish92:
# Overview

# Test Plan

### DIFF
--- a/android/src/main/java/cl/json/RNShareModule.java
+++ b/android/src/main/java/cl/json/RNShareModule.java
@@ -24,6 +24,7 @@ import cl.json.social.GenericShare;
 import cl.json.social.GooglePlusShare;
 import cl.json.social.ShareIntent;
 import cl.json.social.TargetChosenReceiver;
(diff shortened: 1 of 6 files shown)


mikehardy (changes requested):
Hi there! Very cool.

Seems like this needs a related documentation update https://github.com/react-native-share/react-native-share/blob/master/website/docs/share-single.mdx#supported-applications

Any plans on implementing iOS so it's cross-platform?

harish92:
Thanks, 
Sorry, no plans yet implementing on IOS. Sure will do it in the future.

mikehardy (approved):
Looks good to me - and re: iOS, adding support there would be fabulous but I am always a fan of incremental progress (as long as limitations are documented, which they are now - thank you!) so that wouldn't stop me from merging. @MateusAndrade should hit the merge button though I think lest I have missed something

MateusAndrade:
Nice! Everything looks great. Thanks a lot @harish92! 🚀

PR #1060, opened 2021-07-29
Rename Utils to RNShareUtils, to prevent incompatibilites with other libs

### LINKED ISSUE #913: incompatable together 2 libs
I think it happenes because we have put the same class implementation into two different files
Снимок экрана 2020-11-19 в 21 35 09
Also there a same question [https://stackoverflow.com/questions/64856494/2-duplicate-symbols-for-architecture-arm64#](url)

If change react-native-share version from 4.1.0 -> 4.0.4 everything is fine
([https://github.com/react-native-share/react-native-share/compare/v4.0.4...v4.1.0]())
### Steps to reproduce
1. install latest versions of react-native-share (4.1.0) and react-native-sensors (7.0.2)
2. build ios app

### Expected behaviour
App is builded

### Actual behaviour
Error happenes on last build steps

### Platform
iOS

### react-native-share
**Version**: 4.1.0

https://github.com/react-native-share/react-native-share

m-s-will:
# Overview
Simply renames all usages of Utils to RNShareUtils, this should fix #913.


mikehardy (approved):
This is actually a really big deal, I approve of this completely and really appreciate you took the time to do a PR

@MateusAndrade I'm not sure if this merits a breaking release or not. Technically, the public native APIs of a react-native module are a public API surface area as well, including their symbol names, and this *could* cause a compile break for people. It's a matter of taste though. In my opinion version numbers are free, so I'd commit it with a squash-merge commit message like https://docs.google.com/document/d/1QrDFcIiPjSLDn3EL15IJygNPiHORgU1_OOAqWjiDU5Y/edit#heading=h.gbbngquhe0qa

title: 
```
fix(ios, symbols): rename Utils to RNShareUtils to avoid symbol collisions
```

message:

```

BREAKING CHANGE: 

This will probably not affect you, however if you were using the native symbols from 'Utils.h' and 'Utils.m' on iOS directly in Objective-C code, you will need to update your references to the new names 'RNShareUtils.h' and 'RNShareUtils.m'
```

I *always* have to look up exactly how to semantically release a breaking change, and I did so again now - if you're really good at it and this is boring sorry but if it helps, awesome :-)

MateusAndrade:
> I _always_ have to look up exactly how to semantically release a breaking change, and I did so again now - if you're really good at it and this is boring sorry but if it helps, awesome :-)

Thanks for the suggestion. I've been using a similar approach, but this one it far more elegant than mine since we specify what the change can break. 🚀 🚀 🚀 🚀

MateusAndrade (approved):

MateusAndrade:
Just scheduled a `squash merge` with the recommend message @mikehardy. Also, thanks a lot @m-s-will! ❤️

mikehardy:
Yeah - I've become really sensitive as I maintain more repos - to wanting to have the exact change and the actual migration path right there in the changelog if at all possible, so I try to do the same now as a maintainer. But then I forget exactly how every time and have to re-read it. This changelog came out looking beautiful, so it worked :-) :rocket:

PR #1061, opened 2021-07-29
Add shareSingle telegram (iOS only)

akinncar:
# Overview
Add telegram shareSingle to iOS.
As I saw, Telegram share on iOS only allows sharing a simple text or simple text + url.

References:
- https://core.telegram.org/widgets/share
- https://stackoverflow.com/questions/31356360/share-a-link-via-url-scheme-to-telegram

I needed to add a new devDependency to can run commit linter when I changed the example app.

# Test Plan
```jsx
const shareToTelegram = async () => {
  const shareOptions = {
    message: 'Example Telegram',
    url: 'https://google.com',
    social: Share.Social.TELEGRAM,
  };

  try {
    const ShareResponse = await Share.shareSingle(shareOptions);
  } catch (error) {
  }
};
```

# Related
https://github.com/react-native-share/react-native-share/pull/1058


mikehardy (changes requested):
Seems all good to me - except it should include an update to the docs entry as well?

https://github.com/react-native-share/react-native-share/blame/master/website/docs/share-single.mdx#L83

akinncar:
I already did here https://github.com/react-native-share/react-native-share/pull/1061/files#diff-2b156530ac8480a0ede43b606f2c052bea0d549547de15f086317ae5f1168fe4

Am I missing anything?

mikehardy (approved):
:facepalm: how did I miss that? There is nothing missing then, this looks like good work. Thanks!

MateusAndrade:
Thanks @akinncar ! 😄

PR #1069, opened 2021-08-11
title is optional as per documentation

petrbela:


mikehardy:
Just out of curiosity since this formally makes all items optional, are there cases where they are indeed non-optional? Are those cases validated in any way? (like with a sub-type for this thing or the other thing, where it is no longer optional?)

petrbela:
Actually, it should only be optional in the `ShareSingle` interface. I'll fix it.

mikehardy (approved):
works for me, thanks for the extra info

petrbela:
@demedos I think it might be optional but I'm not using it so didn't have enough data to propose that change (I'm not a maintainer). I suggest you create another PR with appropriate changes if you think that's how it should be.

demedos:
done. see https://github.com/react-native-share/react-native-share/pull/1110

PR #1077, opened 2021-08-29
Specify PendingIntent.FLAG_IMMUTABLE for Android S+

robertying:
# Overview

https://github.com/react-native-share/react-native-share/issues/1073 describes the issue.

Ref: Android 12 behavior changes https://developer.android.com/about/versions/12/behavior-changes-12#pending-intent-mutability

# Test Plan
Apply the fix, and the project which targets Android 30 can successfully call `Share.open` without throwing.


MateusAndrade:
Nice! Thanks. 🚀

MateusAndrade (approved):

mikehardy (commented):
This is unfortunately not okay

mikehardy:
https://developer.android.com/reference/android/app/PendingIntent#FLAG_IMMUTABLE was added in API23, which is above the minSdkVersion of this package and of react-native

This needs an API guard or it will crash older APIs.

robertying:
@mikehardy I did use an API level check. What should be the preferred way to do that instead?

mikehardy:
Oh I was reading to fast! I am so sorry @robertying - you did, I see it, and the ternary structure you used should work. Apologies

Issue #1059, opened 2021-07-21
[Android] Share.open always return success: false & dissmissedAction: true


donatoaguirre24:
### Steps to reproduce
1. Use `Share.open` with android
2. Set `failOnCancel: false`

### Expected behavior
The promise resolves if the user shares or dismisses, and the returned object contains `success` and `dismissedAction` properly set.

### Actual behavior
The returned object always contains `success: false` and `dismissedAction: true` no matter if the user shared or dismissed the share sheet.

### Environment
- **React Native version**: 0.64.2
- **React Native platform + platform version**: Android 11

### react-native-share
**Version**: 6.4.0 (latest)

### Repro

```js
const handleShare = async () => {
  try {
    const { success } = await Share.open({ failOnCancel: false, title, url });
    // success always returns false and dissmissedAction always returns true (Android only)
    if (success) {
      Toast.show({ text1: strings.shareSuccess, type: 'info' });
    }
  } catch {
    Toast.show({ text1: strings.shareError, type: 'error' });
  }
};
```

I also tried without using `failOnCancel: false` but then the issue #784 happens :(

kaioduarte:
I investigated this, but I'm not sure how to fix it.

This also happens with `shareSingle` when the `forceDialog` flag is true because we try to use the `IntentSender` from the `TargetChosenReceiver` class (code references for [shareSingle ](https://github.com/react-native-share/react-native-share/blob/master/android/src/main/java/cl/json/social/ShareIntent.java#L252-L255)and [open](
https://github.com/react-native-share/react-native-share/blob/master/android/src/main/java/cl/json/social/SingleShareIntent.java#L78-L82
)).

However, the `RESULT_CANCELED` we check (code [reference](https://github.com/react-native-share/react-native-share/blob/master/android/src/main/java/cl/json/RNShareModule.java#L45-L49)) is not accurate accordingly to the [Android docs](https://developer.android.com/reference/android/app/Activity#startActivityForResult(android.content.Intent,%20int)):

>  if the activity explicitly returned that, didn't return any result, or crashed during its operation.

I wonder if we can send 'success' at that point as the [`expo-sharing`](https://github.com/expo/expo/blob/master/packages/expo-sharing/android/src/main/java/expo/modules/sharing/SharingModule.java#L143-L149) and the [React Native](https://github.com/facebook/react-native/blob/1465c8f3874cdee8c325ab4a4916fda0b3e43bdb/Libraries/Share/Share.js#L46
[...]

PR #1078, opened 2021-08-31
Resolve Promise on successful share for Android

juliar:
# Overview

Before this change, the Promise returned by calling `Share.open()` on Android would not resolve after a successful share, only if sharing was cancelled. Therefore, clients of the library could not handle completion of sharing, such as by removing a loading indicator. With this change, `Share.open()` on Android properly calls the success callback and resolves the Promise.

# Test Plan

Tested with our app's usage of react-native-share. Before this change, on Android, the Promise returned from `Share.open()` would not resolve after sharing the content, and so we wouldn't run the following line of code updating our `loading` state to `false`. With this change, `await Share.open(...)` would complete, the following line of code updating the `loading` state would run, and the loading indicator would be removed.

Let me know if you'd like me to test this in some other way as well.

### DIFF
--- a/android/src/main/java/cl/json/RNShareModule.java
+++ b/android/src/main/java/cl/json/RNShareModule.java
@@ -43,8 +43,12 @@ public class RNShareModule extends ReactContextBaseJavaModule implements Activit
 
     // removed @Override temporarily just to get it working on different versions of RN
     public void onActivityResult(int requestCode, int resultCode, Intent data) {
-        if (requestCode == SHARE_REQUEST_CODE && resultCode == Activity.RESULT_CANCELED) {
-            TargetChosenReceiver.sendCallback(true, false, "CANCELED");
+        if (requestCode == SHARE_REQUEST_CODE) {
+            if (resultCode == Activity.RESULT_CANCELED) {
+                TargetChosenReceiver.sendCallback(true, false, "CANCELED");
+            } else if (resultCode == Activity.RESULT_OK) {
+                TargetChosenReceiver.sendCallback(true, true);
+            }
         }
     }
 


MateusAndrade (approved):
Nice catch! Thanks. 🚀

donatoaguirre24:
@MateusAndrade does this solves #1059 ?

MateusAndrade:
> @MateusAndrade does this solves #1059 ?

I didn't had time to test, would you mind checking this for us? 😄

donatoaguirre24:
> > @MateusAndrade does this solves #1059 ?
> 
> I didn't had time to test, would you mind checking this for us? 😄

Of course, I'll test tomorrow and let you know 👍🏻

donatoaguirre24:
> > @MateusAndrade does this solves #1059 ?
> 
> I didn't had time to test, would you mind checking this for us? 😄

@MateusAndrade sadly it didn't solve the issue 😢

damiandizeo:
It doesn't work for iOS too, even if shared success return false

mikehardy:
@sanduluca did you read #1059 ? I do not believe you should rely on the false/true indicator at all on android. All the other libraries that handle sharing appear to always resolve success and semantically it just means "we successfully opened the share sheet", there appears to be no way to reliably tell if the user really shared or not.

So in your code I would take calling the API to have really only two possible logic branches: it either throws an exception, or it resolves normally (with true or false, does not matter), and that is all of the meaningful information the platform API really offers.

sanduluca:
> 

Yes. I did read it.
I will reformulate my comment
Throws an exception when shared with success

mikehardy:
That is not my experience in my work app on ios or android :thinking: 
https://stackoverflow.com/help/minimal-reproducible-example on a new issue

cooltvglobal:
> Still return false when shared success on android.

same issue

mikehardy:
@cooltvglobal I haven't seen a PR from you proposing a fix - feel free to post one we can merge and release quickly

jamieGardyn:
FYI - I understand there is no fix to this issue at the moment, just wanted to share what I am doing and why.

I am choosing version 5.3.0 of react-native-share (an older version I had been using) which resolves true if the user chooses an app to share to. It doesn't wait until the user finishes the share in order to be certain they really shared (on Android, on iOS it works properly), but I would rather have a false success (which is more likely), than a false error.  And we still get the error if they close the share dialog without choosing an app.

mikehardy:
@jamieGardyn please propose a PR with "false" success (really, altering it on android to simply mean "the intent was *sent* correctly" and we can integrate

jiyuan12354:
same issue, still no fix

PR #1085, opened 2021-09-07
feat: IG android - open share screen

ddereszewski:
# Overview
This introduces a new feature to IG single share. 
I found out how to open share views on IG without providing media.  

# Test Plan
As added in docs

### DIFF
--- a/android/src/main/java/cl/json/social/InstagramShare.java
+++ b/android/src/main/java/cl/json/social/InstagramShare.java
@@ -5,6 +5,7 @@ import android.content.Intent;
 import java.io.File;
 import android.os.Environment;
 import android.net.Uri;
+import android.util.Log;
 
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReadableMap;
@@ -25,13 +26,13 @@ public class InstagramShare extends SingleShareIntent {
     public void open(ReadableMap options) throws ActivityNotFoundException {
         super.open(options);
         try {
-            // passing instagram://share as url option opens Camera View for publishing
+            // passing instagram://share as url option opens Publishing screen with camera view
             if (ShareIntent.hasValidKey("url", options) && options.getString("url").startsWith("instagram://")) {
                 this.getIntent().setAction(Intent.ACTION_VIEW);
                 this.getIntent().setData(Uri.parse(options.getString("url")));
             }
-        }catch (Exception e){
-            // bad param passed
+        } catch (Exception e) {
+            Log.w("RNShare", e.getMessage());
         }
         this.openIntentChooser();
     }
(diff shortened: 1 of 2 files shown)


android/src/main/java/cl/json/social/InstagramShare.java
```diff
@@ -24,7 +24,15 @@ public InstagramShare(ReactApplicationContext reactContext) {
...
     public void open(ReadableMap options) throws ActivityNotFoundException {
         super.open(options);
-        //  extra params here
+        try {
+            // passing instagram://share as url option opens Camera View for publishing
+            if (ShareIntent.hasValidKey("url", options) && options.getString("url").startsWith("instagram://")) {
+                this.getIntent().setAction(Intent.ACTION_VIEW);
+                this.getIntent().setData(Uri.parse(options.getString("url")));
+            }
+        }catch (Exception e){
```
mikehardy, line 33: Best to match existing spacing style

```suggestion
        } catch (Exception e) {
```

android/src/main/java/cl/json/social/InstagramShare.java
```diff
@@ -24,7 +24,15 @@ public InstagramShare(ReactApplicationContext reactContext) {
...
+            // bad param passed
```
mikehardy, line 34: If these are simply consumed with no output whatsover, when the params required inevitably change upstream there will be no notice for anyone anywhere. So these should be logged somehow. The question is severity. For severity "error" means execution should halt as state is undefined - that is not correct here, "info" means no problem really just something you should know - that is also not quite right, "warning" means something unexpected and bad but we think we can continue. I think that fits.

```suggestion
            // bad param passed
            android.util.Log.warn("RNShare", e.getMessage());
```

mikehardy (changes requested):
Interesting - thanks for posting this!

Seems good but spacing doesn't match project and I worry about just eating that exception... what do you think of my suggestions?

ddereszewski:
pushed the fixed. 

I decided to allow `instagram://*` patter for Android as it works but it's nowhere documented. Folks might use it in some other way to achieve some other results. 

Also, I didn't want to introduce any other new param,  increase complexity and make params less generic (Story case). Therefore I suggest a solution to use URL

MateusAndrade (approved):
Nice! Thanks a lot @ddereszewski! 🚀

MateusAndrade:
Do you think we can follow with this @mikehardy? 🤔

mikehardy (approved):
Sure seems fine by me - were suggestions were more sort of general and not related to the feature, I'll defer to anyone (which is certainly @ddereszewski here...) that's actually working on and testing it. I have no idea how Instagram works at an integration / API level haha

PR #1097, opened 2021-09-21
Set `LinkPresentation` as weak framwork on IOS

### LINKED ISSUE #1095: Startup crash on iOS 12 when using Xcode 13
### Steps to reproduce
1. Build and run on iOS 12 device

### Expected behaviour
Works without crashing

### Actual behaviour
Crash on app startup
`dyld: Library not loaded: /System/Library/Frameworks/LinkPresentation.framework/LinkPresentation`

### Environment
- **React Native version**: 0.65.1
- **React Native platform + platform version**: iOS 12.4
- **Typescript version** (if using typescript): 4.4.3

### react-native-share
**Version**: 7.1.0

None

chungweileong94:
# Overview
Since `LinkPresentation` API is only available on iOS 13 and above, we need to mark the `LinkPresentaion` as a weak framework. Else, there will be a startup crash on iOS 12 instead of a build error `dyld: Library not loaded: /System/Library/Frameworks/LinkPresentation.framework/LinkPresentation`. 

Fixes #1095 

# Test Plan
Build the app on iOS 12 device, before the fix the app will crash on startup. And the app should work normally after the fix.

### DIFF
--- a/RNShare.podspec
+++ b/RNShare.podspec
@@ -15,4 +15,6 @@ Pod::Spec.new do |s|
   s.source_files  = "ios/**/*.{h,m}"
 
   s.dependency "React-Core"
+
+  s.ios.weak_framework = 'LinkPresentation'
 end


mikehardy:
Oh I just see the PR now - it wasn't cross linked. You should try this - edit the description and instead of just putting `#issuenumber` do `Fixes #issuenumber` then examine the cross-linking, it's a great github feature

mikehardy (approved):
This seems fine to me and has prior success with a Flutter plugin using the same trick, as cross-linked in the mentioned issue here in this repo

chungweileong94:
> Oh I just see the PR now - it wasn't cross linked. You should try this - edit the description and instead of just putting `#issuenumber` do `Fixes #issuenumber` then examine the cross-linking, it's a great github feature

Oh I see, thx for the tip, as I still fairly new to the Github contribute.

PR #1110, opened 2021-10-06
chore: makes title optional on the ShareOptions

demedos:
# Overview

The title parameter is not required in order for the share intent to properly work

# Test Plan


MateusAndrade (approved):
Nice!! Thanks a lot @demedos!  🚀

PR #1124, opened 2021-11-17
feat(android): add  `useInternalStorage` share option to avoid external storage

xavier-villelegier:
# Overview
Hey hey !
We've noticed that this library is using `getExternalCacheDir`, which usually returns `/sdcard/Android/data/com.package.name/cache`. The thing is, on Android <= 10, this directory can be accessed by any application that has the `WRITE_EXTERNAL_STORAGE` permission, and this can basically lead to any malicious application accessing the previous files you shared. In our case, we share sensitive information, and we don't want them to leak.

I added a `useInternalStorage` option that is replacing it by `getCacheDir` which returns `/data/data/com.package.name/cache`, that can only be accessed by the application with the `com.package.name` identifier.

# Test Plan
I tested it locally and it works fine. My flow was:
- Share a file
- Check that `/sdcard/Android/data/com.package.name/cache` is still empty
- Check that the file is stored in `/data/data/com.package.name/cache`
- Check that the file is correctly shared

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  react-native: react-native-community/react-native@5.3.0
+  react-native: react-native-community/react-native@6.0.0
 
 commands:
   checkout-attach-workspace:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  react-native: react-native-community/react-native@6.0.0
+  react-native: react-native-community/react-native@6.0.1
 
 commands:
   checkout-attach-workspace:
--- a/android/src/main/java/cl/json/RNSharePathUtil.java
+++ b/android/src/main/java/cl/json/RNSharePathUtil.java
@@ -76,7 +76,7 @@ public class RNSharePathUtil {
                 final String type = split[0];
 
                 if ("primary".equalsIgnoreCase(type) || "0".equalsIgnoreCase(type)) {
-                    return filePrefix + context.getExternalCacheDir() + "/" + split[1];
+                    return filePrefix + context.getCacheDir() + "/" + split[1];
                 } else if ("raw".equalsIgnoreCase(type)) {
                     return filePrefix + split[1];
                 } else if (!TextUtils.isEmpty(type)) {
--- a/android/src/main/java/cl/json/RNSharePathUtil.java
+++ b/android/src/main/java/cl/json/RNSharePathUtil.java
@@ -76,7 +76,7 @@ public class RNSharePathUtil {
                 final String type = split[0];
 
                 if ("primary".equalsIgnoreCase(type) || "0".equalsIgnoreCase(type)) {
-                    String cacheDir = useInternalStorage ? context.getCacheDir() : context.getExternalCacheDir()
+                    File cacheDir = useInternalStorage ? context.getCacheDir() : context.getExternalCacheDir();
                     return filePrefix + cacheDir + "/" + split[1];
                 } else if ("raw".equalsIgnoreCase(type)) {
                     return filePrefix + split[1];
--- a/android/src/main/java/cl/json/RNSharePathUtil.java
+++ b/android/src/main/java/cl/json/RNSharePathUtil.java
@@ -63,7 +63,7 @@ public class RNSharePathUtil {
         return result;
     }
 
-    public static String getRealPathFromURI(final Context context, final Uri uri) {
+    public static String getRealPathFromURI(final Context context, final Uri uri, Boolean useInternalStorage) {
 
         String filePrefix = "";
         // DocumentProvider
@@ -76,7 +76,8 @@ public class RNSharePathUtil {
                 final String type = split[0];
 
                 if ("primary".equalsIgnoreCase(type) || "0".equalsIgnoreCase(type)) {
-                    return filePrefix + context.getCacheDir() + "/" + split[1];
+                    String cacheDir = useInternalStorage ? context.getCacheDir() : context.getExternalCacheDir()
+                    return filePrefix + cacheDir + "/" + split[1];
                 } else if ("raw".equalsIgnoreCase(type)) {
                     return filePrefix + split[1];
                 } else if (!TextUtils.isEmpty(type)) {
--- a/android/src/main/java/cl/json/ShareFile.java
+++ b/android/src/main/java/cl/json/ShareFile.java
@@ -111,7 +111,7 @@ public class ShareFile {
         return this.type;
     }
     private String getRealPathFromURI(Uri contentUri) {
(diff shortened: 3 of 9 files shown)


mikehardy:
Hi there!
Thanks for taking the time to post this!

1) This is not a regression as implied in comment - it was the case in all android versions, just now with scoped storage they are putting out a lot of documentation about it. The external storage directories have always been wide open to apps with that permission, and that permission used to be very freely granted. So this is an existing problem for you on all APIs, not just Android 10

That's just commentary thought - a technical item

2) this is an unwanted breaking change, apps may be relying on external storage in order to share very large files from very inexpensive phones with small internal storage that still allow expansion cards. So this won't "protect" them, it will make their use case potentially impossible. My target market for instance is dominated by this type of phone so I am sensitive to this

However, if you added an optional parameter to the affected APIs "secureIfPossible" or something (I just made that up) and passed that through to native where you used that to determine which cache dir to use, that could work.

In the meantime, you may always use something like https://github.com/itinance/react-native-fs/#api for your needs and share via base64 strings?

xavier-villelegier:
Hey @mikehardy, thanks for the fast feedback ⚡ 

> This is not a regression as implied in comment 

I was talking about Android **10 and below**, not a regression, but a behavior that has been fixed on Android 11 (https://developer.android.com/about/versions/11/privacy/storage), now the external storage is also scoped, that was not the case before (so we're trying to fix the issue for Android 10 and below here)

> However, if you added an optional parameter to the affected APIs

Would a bool parameter `useInternalStorage` that is false by default work for you ?

mikehardy:
I can't speak for @MateusAndrade who is the normal merge+release person, but that would make it non-breaking, assuming it was an optional parameter, so yes

On higher APIs there are many apps that still have storage management permissions, so external storage can still not be considered safe. Honestly if security were a concern it seems the data should not hit disk at all, or should not do so without being encrypted somehow - the same thing might apply to iOS if the platform is compromised at all (which is apparently not hard and happens to "people of interest" all the time, reference NSO group, Pegasus et al). I guess what I'm getting at is that this decreases attack surface area, which is a good thing and I am all for incremental progress and defense in depth, but it may be a false sense of security depending on your project's threat sensitivity

xavier-villelegier:
> Honestly if security were a concern it seems the data should not hit disk at all

100% agree on that. And that's actually one of the questions I had: was it a technical choice to download the file into the cache to share it ? Or is it by design on Android ? I tried to share files on other apps like Gmail and it looks like files are downloaded in the internal storage, even if an sdcard is installed. So the best thing we can think of right now is to at least move files in the internal storage that can't be accessed by other apps, but of course in term of security, it's not unbreakable

mikehardy:
I'm unsure on the exact history of the decision to use external, I was only analyzing in terms of change impact. I do know some people share large PDFs based on issue traffic here, unsure how gmail handles attachments that fill internal storage, as would happen nearly immediately on many of the phones here in Ecuador (people literally uninstall and reinstall apps from their "core app set" just because they don't all fit!).

xavier-villelegier:
Hey @mikehardy @MateusAndrade, it should be ready to review

I decided to put it in the `ShareFile` and `ShareFiles` classes, otherwise we'd have to pass it through a lot of levels and functions to the `getURI` function

I tested locally and without the option the file is downloaded `/sdcard/Android/data/com.package.name/cache`, and when you pass `useInternalStorage: true`, the file is downloaded in `/data/data/com.package.name/cache`

Let me know if it's good for you 👌

mikehardy:
How unfortunate that the react-native-community CircleCI orb was using a combination of a JDK8-specific JVM option and an unpinned Android Docker image that updated to JDK11 underneath them/us. So our Android build CI is busted at the moment, see:

https://github.com/react-native-community/react-native-circleci-orb/issues/120
https://github.com/react-native-community/react-native-circleci-orb/pull/121
https://github.com/react-native-community/react-native-circleci-orb/pull/122

mikehardy (dismissed):
This looks good to me, I left a note about the CI failure
I've been validating builds on other projects all day long and don't have time to check this right now unfortunately
Can you verify that if you depend on the commit hash from your PR in the package.json on your project this works in your project? I have no idea how long it will take the CircleCI orb repair PRs to merge so this might be the fastest way to go - can always fix and release a fix if needed after

xavier-villelegier:
> Can you verify that if you depend on the commit hash from your PR in the package.json on your project this works in your project?

Hey @mikehardy, thanks for the approval 👌 

I can confirm that my projects still work on this commit hash. ✅ 

What I tested:
- Without the option, it work as before, and the file is downloaded on `/sdcard`
- With the new `useInternalStorage` option, nothing is downloaded in `/sdcard`, everything goes in `/data`, and the file work as expected once it's shared

mikehardy:
Great! My PRs to the react-native-community circleci orb were actually approved overnight as well, I'm trying to test them here (a: to make sure the changes were actually good! and b: just as good CI hygiene) but if that doesn't work out it still seems good to merge, in my opinion, thanks again @xavier-villelegier

mikehardy:
I fixed up all the react-native-community CI orbs we use for CI here, and integrated them in to this PR's fork, they ran and succeeded here

https://github.com/react-native-community/react-native-circleci-orb/pull/123#issuecomment-974095135

@MateusAndrade this is ready for review and merge if you like I think?

MateusAndrade:
> I fixed up all the react-native-community CI orbs we use for CI here, and integrated them in to this PR's fork, they ran and succeeded here
> 
> [react-native-community/react-native-circleci-orb#123 (comment)](https://github.com/react-native-community/react-native-circleci-orb/pull/123#issuecomment-974095135)
> 
> @MateusAndrade this is ready for review and merge if you like I think?

Wow! Sorry for the long time to answer, I've been quite busy in the last weeks. 😅 

Do you guys think we should merge this as a major release or we are good to go with a minor/feature?

MateusAndrade (approved):

mikehardy:
It looked to me like it was fully backwards compatible - I edited the title to reflect it

MateusAndrade:
@xavier-villelegier, @mikehardy done! We should have a new release in some minutes! 🔥

mikehardy:
boom! :rocket:  and react-native-share, if used carefully, just got a little bit more secure, making the world a better place :smile:

xavier-villelegier:
Thank you guys for your reactivity, you rock ! It's available ! 🔥

merlecap3:
Thank you for the great feature! But the new property "useInternalStorage" is missing in the typings at types.ts. Can anyone add the new property?

MateusAndrade:
> Thank you for the great feature! But the new property "useInternalStorage" is missing in the typings at types.ts. Can anyone add the new property?

Nice catch! I think this can be solved by adding this type here: https://github.com/react-native-share/react-native-share/blob/400f1d188cdeb6902b9bfcee0d9517be7bf55d6a/src/types.ts#L88-L104

PR #1126, opened 2021-11-24
fix(instagram): fix black screen issue for instagram stories with Social.Instagram

alexisrougnant:
# Overview
Sharing to Instagram using `shareSingle` and the `social` property `Social.Instagram` opens an intent chooser that lets the user decide between sharing as Instagram Direct, Feed, or Stories. Direct and Feed are working but Stories doesn't.
The issue looks similar to [this one](https://github.com/react-native-share/react-native-share/issues/1084)

For sharing as Stories, instead of using the Android intent action `ACTION_SEND`, one needs to use the action ` "com.instagram.share.ADD_TO_STORY"` as in the Instagram documentation [example](https://developers.facebook.com/docs/instagram/sharing-to-stories).

To fix the issue - and only for image data - we create a chooser intent that combines the `ACTION_SEND` intent and the Instagram `ADD_TO_STORY`  intent as extra initial intent. The approach is based on [this gist](https://gist.github.com/michaeltys/a8613e5aea9db8e4684bf85568e40160).

# Test Plan
Use `shareSingle` function with the `Social.Instagram` `social` property.

1. with a scheme url, like ` 'instagram://share'` => should not have changed
2. with image data => the black screen issue has been fixed
3. with another data  => should not have changed

### DIFF
--- a/android/src/main/java/cl/json/social/InstagramShare.java
+++ b/android/src/main/java/cl/json/social/InstagramShare.java
@@ -58,7 +58,8 @@ public class InstagramShare extends SingleShareIntent {
     }
 
     protected void openInstagramIntentChooserForImage(String url, String chooserTitle) {
-        ShareFile shareFile = new ShareFile(url, "image/jpeg", "image" , this.reactContext);
+        Boolean shouldUseInternalStorage = ShareIntent.hasValidKey("useInternalStorage", options) && options.getBoolean("useInternalStorage");
+        ShareFile shareFile = new ShareFile(url, "image/jpeg", "image", shouldUseInternalStorage, this.reactContext);
         Uri uri = shareFile.getURI();
 
         Intent feedIntent = new Intent(Intent.ACTION_SEND);
--- a/android/src/main/java/cl/json/social/InstagramShare.java
+++ b/android/src/main/java/cl/json/social/InstagramShare.java
@@ -32,7 +32,13 @@ public class InstagramShare extends SingleShareIntent {
                 return;
             }
             String url = options.getString("url");
+
             Boolean isUrlScheme = url.startsWith("instagram://");
+            if (isUrlScheme) {
+                openInstagramUrlScheme(url);
+                return;
+            }
+
 
             if (!ShareIntent.hasValidKey("type", options)) {
                 Log.e("RNShare", "No type provided");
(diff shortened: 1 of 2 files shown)


MateusAndrade (commented):
Nice! Do you think we could improve our docs regarding this change? Maybe here: https://react-native-share.github.io/react-native-share/docs/share-single

alexisrougnant:
> Nice! What do you think we could improve our docs regarding this change? Maybe here: https://react-native-share.github.io/react-native-share/docs/share-single

alexisrougnant:
> Nice! What do you think we could improve our docs regarding this change? Maybe here: https://react-native-share.github.io/react-native-share/docs/share-single

I updated the `singleShare` documentation. 

(Sorry for closing the PR, it was not intentional)

MateusAndrade (approved):

MateusAndrade:
> > Nice! What do you think we could improve our docs regarding this change? Maybe here: https://react-native-share.github.io/react-native-share/docs/share-single
> 
> I updated the `singleShare` documentation.
> 
> (Sorry for closing the PR, it was not intentional)

No problem, can you send a commit to force a new build at CircleCI?

alexisrougnant:
> > > Nice! What do you think we could improve our docs regarding this change? Maybe here: https://react-native-share.github.io/react-native-share/docs/share-single
> > 
> > 
> > I updated the `singleShare` documentation.
> > (Sorry for closing the PR, it was not intentional)
> 
> No problem, can you send a commit to force a new build at CircleCI?

It doesn't seem to have trigger the CI @MateusAndrade

lordkiz:
The blank screen issue for Instagram Stories is still happening, unfortunately.
Android 10.

NikoFarrelly:
> The blank screen issue for Instagram Stories is still happening, unfortunately. Android 10.

I've run into the same issue, the same solution from the initial thread reporting the problem has solved it for me. 

> I managed to get it working by specifing manually the extyension when creating the background SharedFile class:
> 
> I edit this:
> 
> `ShareFile backgroundAsset = new ShareFile(backgroundFileName, "background" , this.reactContext);`
> 
> to this:
> 
> `ShareFile backgroundAsset = new ShareFile(backgroundFileName, "image/jpeg", "background" , this.reactContext);`
> 
> For some reason, it was trying to create a file with a null extension when I tried to specify the background file as base64.


https://github.com/react-native-share/react-native-share/issues/1084#issuecomment-932348328

mikehardy:
Does this need a PR in order to allow specification of the mime type? Or is it *always* `image/jpeg` and this change is general to all share file backgrounds and we may hard code it, or could we do mime type detection of the file?

NikoFarrelly:
> 

I've just pushed up a PR which has the hardcoded mime type, I assume somebody has a use case for a different file type (other than jpeg). So mime type detection would be great in the long-term for this library. I'd include it in my PR, but unfortunately I'm a RN dev, not an android dev 😅

PR #1137, opened 2022-01-12
Fixed android instagram stories black screen

NikoFarrelly:
Solution found by https://github.com/marf on this thread https://github.com/react-native-share/react-native-share/issues/1084#issuecomment-932348328

# Overview

Kia ora! Currently on the android side of the library, when sharing to an instagram story your image won't appear, and instead you'll see a blackscreen. @marf found a solution, and I've implemented it here. 

There was another PR which aimed to fix this issue https://github.com/react-native-share/react-native-share/pull/1126
However it only updated the `InstagramShare.java`, not the `InstagramStoriesShare.java`, hence this PR.

# Test Plan

Use the `Share.shareSingle` function with the Instagram stories prop, along with a background image.

e.g. `await Share.shareSingle({social: Share.Social.INSTAGRAM_STORIES,
            backgroundImage: image})`

Before this fix you'll be greeted with a black screen
|Before|
|-|
|before|

After this fix you'll see your image in a story
|After|
|-|
|after|

### DIFF
--- a/android/src/main/java/cl/json/social/InstagramStoriesShare.java
+++ b/android/src/main/java/cl/json/social/InstagramStoriesShare.java
@@ -93,7 +93,7 @@ public class InstagramStoriesShare extends SingleShareIntent {
                 backgroundFileName = options.getString("backgroundVideo");
             }
 
-            ShareFile backgroundAsset = new ShareFile(backgroundFileName, "background", useInternalStorage, this.reactContext);
+            ShareFile backgroundAsset = new ShareFile(backgroundFileName, "image/jpeg", "background", useInternalStorage, this.reactContext);
 
             this.intent.setDataAndType(backgroundAsset.getURI(), backgroundAsset.getType());
             this.intent.setFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION);


mikehardy (approved):
You know, strangely the previous PR linked in this one adds a "type" requirement, but then just doesn't use it. It hard-codes `image/jpeg` just like here. So I suppose this is fine (it does the same as the previous PR, which no one is complaining about?) but we could probably use a tweak to the docs here where we tell people that we expect a jpeg file, as that's the mime type we send to instagram - altering the previous documentation bit where we require people to specify the type

examine hard coding here https://github.com/react-native-share/react-native-share/pull/1126/files#diff-f4bb2fed0cda5a56b51eff86a6f6df656221652240d87778270db7c70d8f0333R66

mikehardy:
tēnā rawa atu koe! (ok I had to look that up, but I spent 6 weeks traveling all.over the south island and a bit of the north, and your  Maori greeting brought it back, such great times)

NikoFarrelly:
> You know, strangely the previous PR linked in this one adds a "type" requirement, but then just doesn't use it. It hard-codes `image/jpeg` just like here. So I suppose this is fine (it does the same as the previous PR, which no one is complaining about?) but we could probably use a tweak to the docs here where we tell people that we expect a jpeg file, as that's the mime type we send to instagram - altering the previous documentation bit where we require people to specify the type
> 
> examine hard coding here https://github.com/react-native-share/react-native-share/pull/1126/files#diff-f4bb2fed0cda5a56b51eff86a6f6df656221652240d87778270db7c70d8f0333R66

That's a great catch! You're definitely going to save some hair pulling by updating the documentation to explicitly reference jpgs being required. Adding in mime type detection like you mentioned previously would also be great for the extensibility of this lib 😄 



> tēnā rawa atu koe! (ok I had to look that up, but I spent 6 weeks traveling all.over the south island and a bit of the north, and your Maori greeting brought it back, such great times)

That's so lovely to hear! I'm glad you loved traveling through our country ❤️  Much aroha whanau 

Also thanks for the quick feedback! 10m after the PR is up & it's approved, you're a legend 🤙

MateusAndrade (approved):
Nice catch! Thanks a lot. Also, your PR description looks neat. 🚀 🚀 🚀

marf:
Thank you for making the PR :)

PR #1140, opened 2022-01-17
fix(instagram): Fix black screen when sharing to Instagram Stories with a sticker

Sk92:
# Overview
Ciao! 👋🏻
I would like to improve the PR #1137 (related to #1084) to fix this issue also when the shares includes a sticker image (a png over the background).
The solution is the same implemented in #1137 but it is applied to the "stickerImage" param with the only one difference that a sticker is a png and not a jpg.

The issue is only on Android.

# Test Plan
Use the `Share.shareSingle` function with the `Share.Social.INSTAGRAM_STORIES` prop and a sticker image.
You can use this code below to reproduce the bug:
```
const shareOptions = {
      stickerImage: `data:image/png;base64,${base64Image}`,
      backgroundTopColor: '#15B9DD',
      backgroundBottomColor: '#063478',
      social: Share.Social.INSTAGRAM_STORIES,
};

await Share.shareSingle(shareOptions);
```

Before this fix you'll see only a black screen or the Instagram app that tries to start but immediately crashes.
For the screenshots you could refer to the ones attached to PR #1137

PR #1141, opened 2022-01-17
chore(example): update example to rn66

mikehardy:
# Overview

fixes android build breaks (jcenter is super flaky now, react-native 66 doesn't need it anymore) and fixes ios build breaks (react-native < 0.64.2 does not even build on Xcode > 12.4)

# Test Plan

1. Run example app locally (`yarn && yarn start` in one terminal, `yarn start:android` and/or `yarn start:ios` in another terminal)
1. see the same build error seen in #1140 
1. use upgrade-helper to update from react-native 0.63.x to 0.66.x
1. re-run the example app locally, see it pass


mikehardy:
There were still 2 things to fix it - one was an android cliPath item, and one was cocoapods out of date on CircleCI because we were using old Xcode issue - I just pushed fixes for them - circleci should go green on this one

mikehardy:
:boom: 

image

MateusAndrade (approved):

MateusAndrade:
Nice! I'm dealing with this JCenter problem right now. 😆 Thanks!!! 🚀

mikehardy:
Yeah, I'm disappointed that they let the service degrade after they said they'd keep it up, but :shrug: 
If you update to react-native 0.65+ everything is fine, and they just merged react-native-macos from 0.64 to 0.66 so even macos dependencies don't have to hold you back now.
I still have one dependency in one project on jcenter but should be able to move it to jitpack...such a pain!

PR #1148, opened 2022-01-28
fix: support unknown whatsapp number

veraldo:
# Overview
Workaround for the issue https://github.com/react-native-share/react-native-share/issues/952 . Images were not being shared to a specific number if the conversation was never opened before. The workaround found is to open a blank conversation first, then trying to share it again.

# Test Plan
On the App.js from example app, change the `shareSingleImage` function to the following:

```javascript
  /**
   * This functions share a image passed using the
   * url param
   */
  const shareSingleImage = async () => {
    const shareOptions = {
      title: 'Share file',
      url: images.image1,
      failOnCancel: false,
      social: 'whatsapp',
      whatsAppNumber:
        'Insert a number never contacted before here. Must include the country code, and not include symbols.',
    };

    try {
      const ShareResponse = await Share.shareSingle(shareOptions); // change it to shareSingle();
    } catch (error) {
    }
  };
```

### DIFF
--- a/android/src/main/java/cl/json/social/WhatsAppBusinessShare.java
+++ b/android/src/main/java/cl/json/social/WhatsAppBusinessShare.java
@@ -13,6 +13,11 @@ public class WhatsAppBusinessShare extends SingleShareIntent {
 
     private static final String PACKAGE = "com.whatsapp.w4b";
     private static final String PLAY_STORE_LINK = "market://details?id=com.whatsapp.w4b";
+    
+    private static final String START_CONVERSATION_CLASS = "com.whatsapp.Conversation";
+    private static final String SHARE_TO_CONVERSATION_CLASS = "com.whatsapp.ContactPicker";
+    
+    private static final int START_ACTIVITY_TIME_GAP_MS = 300;
 
     public WhatsAppBusinessShare(ReactApplicationContext reactContext) {
         super(reactContext);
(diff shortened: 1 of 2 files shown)


android/src/main/java/cl/json/social/WhatsAppBusinessShare.java
```diff
@@ -19,6 +20,24 @@ public WhatsAppBusinessShare(ReactApplicationContext reactContext) {
...
+        
+        if (options.hasKey("whatsAppNumber")) {
+            // create an empty conversation in case it's not on contacts
+            this.getIntent().setComponent(new ComponentName(PACKAGE, "com.whatsapp.Conversation")); 
+            this.openIntentChooser();
+
+
+            // leave room for the conversation to be created
+            try {
+                Thread.sleep(300);
```
MateusAndrade, line 32: What about setting this `300` to a constant? It would be much easier to change/find what this number does.

android/src/main/java/cl/json/social/WhatsAppShare.java
```diff
@@ -19,6 +20,24 @@ public WhatsAppShare(ReactApplicationContext reactContext) {
...
+        
+        if (options.hasKey("whatsAppNumber")) {
+            // create an empty conversation in case it's not on contacts
+            this.getIntent().setComponent(new ComponentName(PACKAGE, "com.whatsapp.Conversation")); 
+            this.openIntentChooser();
+
+
+            // leave room for the conversation to be created
+            try {
+                Thread.sleep(300);
```
MateusAndrade, line 32: We can apply the same idea about the `constant` here. Do you think it would be a good idea to share this number between the two modules?

MateusAndrade (changes requested):
Nice catch! What do you think about adding these changes to our docs?

MateusAndrade (approved):

PR #1152, opened 2022-02-02
Use root locale when changing case in java

halaei:
Not setting locale for language/country neutral operation may cause bug depending on the default locale.
See https://docs.oracle.com/javase/7/docs/api/java/util/Locale.html#ROOT

Note: I am just searching for toLowerCase() and toUppercase() in my project's dependencies and send the same PR, in order to potentially raise the awareness. Although I've seen the lack of explicit locale has caused issues for us, I am not sure if react-native-share is actually affected.

Example related issue: joltup/rn-fetch-blob#573

# Overview

# Test Plan

### DIFF
--- a/android/src/main/java/cl/json/RNShareModule.java
+++ b/android/src/main/java/cl/json/RNShareModule.java
@@ -14,6 +14,7 @@ import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.bridge.Callback;
 
 import java.util.HashMap;
+import java.util.Locale;
 import java.util.Map;
 
 import cl.json.social.EmailShare;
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Great catch! This has bitten me specifically before - it's rare but bites hard when it gets you 
https://github.com/invertase/react-native-firebase/issues/3917
https://github.com/invertase/react-native-firebase/commit/eb8d89306fd569d7ef64298a99e970c653b79178#diff-908a605f72b07016de78a78bd642e6566f0f949f2ed64190315823d3e45949f2

A full fix here would enable lint in the example app tuned to be fatal for this issue but that's more involved, I won't block the PR on it but it's a strong desire if you've got time to add that @halaei ?

halaei:
@mikehardy Thanks for the links, but I know too little about lint at the moment to do this myself any time soon.

PR #1181, opened 2022-04-14
fix(ios): Replace initWithURLs with initForExportingURLs on iOS 15+

luoxuhai:
# Overview

[initWithURLs:inMode](https://developer.apple.com/documentation/uikit/uidocumentpickerviewcontroller/2921628-initwithurls?language=objc)  deprecated on iOS 15+. Use [initForExportingURLs](https://developer.apple.com/documentation/uikit/uidocumentpickerviewcontroller/3566731-initforexportingurls?language=objc) instead on iOS 15+.

# Test Plan

**Device Test:**

iOS15+:

Mac Catalyst:

### DIFF
--- a/ios/RNShare.m
+++ b/ios/RNShare.m
@@ -246,10 +246,15 @@ RCT_EXPORT_METHOD(open:(NSDictionary *)options
             RCTLogError(@"No `urls` to save in Files");
             return;
         }
-        if (@available(iOS 11.0, *)) {
+        if (@available(iOS 11.0, macCatalyst 13.1, *)) {
             resolveBlock = successCallback;
             rejectBlock = failureCallback;
-            UIDocumentPickerViewController *documentPicker = [[UIDocumentPickerViewController alloc] initWithURLs:urls inMode:UIDocumentPickerModeExportToService];
+            UIDocumentPickerViewController *documentPicker = nil;
+            if (@available(iOS 15.0, macCatalyst 15.0, *)) {
+                documentPicker = [[UIDocumentPickerViewController alloc] initForExportingURLs:urls asCopy:YES];
+            } else {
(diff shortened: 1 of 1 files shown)


ios/RNShare.m
```diff
@@ -249,7 +249,12 @@ - (NSDictionary *)constantsToExport
         if (@available(iOS 11.0, *)) {
             resolveBlock = successCallback;
             rejectBlock = failureCallback;
-            UIDocumentPickerViewController *documentPicker = [[UIDocumentPickerViewController alloc] initWithURLs:urls inMode:UIDocumentPickerModeExportToService];
+            UIDocumentPickerViewController *documentPicker = nil;
+            if (@available(iOS 15.0, *)) {
```
mikehardy, line 253: Just in case people are using this on catalyst too, this is a bit more complete, but it's really the same idea of course

```suggestion
            if (@available(iOS 15.0, macCatalyst 15.0, *)) {
```

mikehardy (changes requested):
Hey, thanks for the PR! This seems like the right thing to do. I proposed a tiny change to make sure catalyst builds will do things correctly too since those are supported out of the box on react-native normally. What do you think?

Also, have you tested this locally on an iOS15+ device and you confirm it works?

mikehardy (approved):
looks good to me!

ios/RNShare.m
```diff
@@ -246,10 +246,15 @@ - (NSDictionary *)constantsToExport
             RCTLogError(@"No `urls` to save in Files");
             return;
         }
-        if (@available(iOS 11.0, *)) {
+        if (@available(iOS 11.0, macCatalyst 13.1, *)) {
```
mikehardy, line 249: Nice touch thank you

mikehardy (approved):
Looks good to me thank you!

PR #1183, opened 2022-04-14
fix(ios): Xcode project Utils to RNShareUtils

### LINKED ISSUE #1180: Utils.m reference in the RNShare.xcodeproj file breaks my build
## Ask your Question
First of all I wanted to say thank you for your amazing work!

Second I wanted to report an issue, which I'm not sure is my fault or a problem with the library. I have an issue with the `Utils.m` file being referenced in the `RNShare.xcodeproj` file. I linked one of the references as an example:
https://github.com/react-native-share/react-native-share/blob/main/ios/RNShare.xcodeproj/project.pbxproj#L10

This breaks my build in Xcode since it tries to access the non-existent Utils.m file. Is there something I missed or is it a problem with the library?

I would highly appreciate your help with that!

mikehardy:
Missed as part of #1060 
Fixes #1180

# Test Plan

@swapzone indicates this works for them locally?

### DIFF
--- a/ios/RNShare.xcodeproj/project.pbxproj
+++ b/ios/RNShare.xcodeproj/project.pbxproj
@@ -7,7 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		4EED9CAF24F5A2EB006505C4 /* Utils.m in Sources */ = {isa = PBXBuildFile; fileRef = 4EED9CAE24F5A2EB006505C4 /* Utils.m */; };
+		4EED9CAF24F5A2EB006505C4 /* RNShareUtils.m in Sources */ = {isa = PBXBuildFile; fileRef = 4EED9CAE24F5A2EB006505C4 /* RNShareUtils.m */; };
 		6D697F8423A86EBC00F754A1 /* RNShareActivityItemSource.m in Sources */ = {isa = PBXBuildFile; fileRef = 6D697F8323A86EBC00F754A1 /* RNShareActivityItemSource.m */; };
 		B33C22851D441713006BCD98 /* GenericShare.m in Sources */ = {isa = PBXBuildFile; fileRef = B33C22841D441713006BCD98 /* GenericShare.m */; };
 		B33C22891D4419E8006BCD98 /* WhatsAppShare.m in Sources */ = {isa = PBXBuildFile; fileRef = B33C22881D4419E8006BCD98 /* WhatsAppShare.m */; };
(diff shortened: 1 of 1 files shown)


swapzone:
Let me check that real quick since I did not modify that file manually but instead changed in in the Xcode UI.

swapzone:
It shows a different error now, could be related to my local setup though:

mikehardy:
So, are you opening the .xcproj for RNShare directly? Perhaps the reason this has never failed for others is that we all use Cocoapods now, and we open the .xcworkspace for the whole app, not the .xcproj. This ios code is only meant to be consumed as an integrated part of the .xcworkspace

Why do I think that might be the case? Because if you can't find NSObject as a symbol nothing is going to work and this would have shown up for others a long time ago. This is all very puzzling, and I can't reproduce it as a problem, even though obviously it should be RNShareUtil instead of Util now inside the pbxproj file

swapzone:
Just to be clear on what I'm doing: That failing build only happens to me when I add the `RNShare.xcodeproj` to my Library folder via the "Add Files to MY_PROJECT_NAME" option as described in the setup instructions:

Screenshot 2022-04-15 at 09 12 05

I'm rather new to React Native, so I might be missing some crucial points here. And if I'm the only one having issues, we should not waste more time on it.

mikehardy:
Oh wow! Yes, there is no need for you to ever manually add any react-native module code to your iOS project
Behold auto-linking! Handled by cocoapods scripts :-)
https://github.com/react-native-community/cli/blob/master/docs/autolinking.md

This PR seems like it may have value but also it is not happening for others so I'm going to go for the "minimal intervention" then which is "no intervention" and close it

That's based on the idea that auto-linking really *should* work and you will be moving forward after stopping with the manual add

petrgazarov:
I'm still getting this error even after auto-linking.
```
Build input file cannot be found: '/Users/XXX/project/node_modules/react-native-share/ios/Utils.m'
```

This is with the latest version of `react-native-share` (7.3.9). Anything obvious I'm missing?

mikehardy:
Whatever is missing is going to be non-obvious, but is *likely* going to involve a non-standard react-native Xcode project setup, vs that which you would get with a fresh `npx react-native init <myprojectname>` + `yarn add react-native-share`

If you did those two steps (with a dummy project name), then a `yarn ios` would work, which shows there is not really a problem, this is just sort of a "canary" problem hinting at a deeper Xcode project  issue

*I think*

PR #1184, opened 2022-04-15
fix(ios): custom title got over written by meta data

Hezron-Nelapati:
# Overview

The title which user has provided in the linkMetaData options should not be overwritten by metadata from url, unless title is not specified by the user.

# Test Plan

Test Case 1:

Pass the title prop in LinkMetaData options, while sharing the title passed should be present even if the title in metadata is different.

Test Case 2:

If title is not specified in the LinkMetaData options, while sharing the title must be taken from metadata in URL

### DIFF
--- a/ios/RNShareActivityItemSource.m
+++ b/ios/RNShareActivityItemSource.m
@@ -66,7 +66,9 @@
             } else {
                 self->linkMetadata.originalURL = metadata.originalURL;
                 self->linkMetadata.URL = metadata.URL;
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
Certainly seems reasonable, thanks for this!

spencerjsmall:
Hi there, I'm getting this issue too with the existing metadata icon overriding the provided one. Any help would be great.

PR #1198, opened 2022-05-07
Fix: remove old style double-quote imports

coolsoftwaretyler:
# Overview

Hey folks, 

This PR closes https://github.com/react-native-share/react-native-share/issues/1197. It seems that in Expo SDK 44, the Expo team made some changes to `React-Core.podspec`, which now break on the old style double-quote imports. [Here is a comment from an issue that very precisely describes the changes and their implications](https://github.com/expo/expo/issues/15622#issuecomment-997225774).

I haven't really written Objective-C since like 2012, so in order to find my way around, I modeled the changes after this `patch-package` diff for a different package: https://github.com/expo/expo/issues/15622#issuecomment-997141629. Mostly I swapped out the conditional imports with the double-quoted `RCT` classes. 

# Test Plan

Once I made my changes, I started the [example application](https://react-native-share.github.io/react-native-share/docs/contributing#using-the-example-app). It built fine for iOS and the buttons fired. I got some errors in the console when I tried things like sharing to Instagram or Telegram, but mostly these errors said things like "not installed", and I'm assuming this is because I didn't have the corresponding apps installed on that particular simulator. 

To check if it was my changes or missing apps on the simulator, I ran the example app from the `main` branch, and I saw the same behavior: IG share and FB share opened an empty Safari tab. 

If there's a better way for me to set up a test environment, will you folks let me know? I didn't see much else in the contribution guide.

### DIFF
--- a/ios/EmailShare.h
+++ b/ios/EmailShare.h
@@ -8,46 +8,16 @@
 
 
 #import <UIKit/UIKit.h>
-// import RCTConvert
-#if __has_include(<React/RCTConvert.h>)
+// import RCTConvertttt
 #import <React/RCTConvert.h>
-#elif __has_include("RCTConvert.h")
-#import "RCTConvert.h"
-#else
-#import "React/RCTConvert.h"   // Required when used as a Pod in a Swift project
-#endif
 // import RCTBridge
-#if __has_include(<React/RCTBridge.h>)
 #import <React/RCTBridge.h>
-#elif __has_include("RCTBridge.h")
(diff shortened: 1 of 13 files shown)


MateusAndrade (approved):

mikehardy (commented):
I'm late to the party here reviewing this - but indeed that old style in the compiler conditional was *quite* old and I've seen this successfully removed in a couple other modules. This should increase compatibility without causing problems. Thanks for posting it @coolsoftwaretyler

PR #1221, opened 2022-06-11
ci(node): setting publish version to 14

MateusAndrade:
# Overview

Setting node version to 14, in order to properly publish new packages due to #1217 bumping semantic-release

# Test Plan

Run at CircleCI 🚀

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,7 +61,9 @@ jobs:
           device: iPhone 11
 
   publish-version:
-    executor: react-native/linux_js
+    executor: 
+      name: react-native/linux_js
+      node_version: '14'
     working_directory: ~/app
     steps:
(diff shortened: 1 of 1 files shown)


mikehardy (commented):
incidentally, why not 16 ? 14 will be EOL almost immediately and while we're in here paying attention / testing it, seems best to move the next update point a year or two farther away ?

MateusAndrade:
> incidentally, why not 16 ? 14 will be EOL almost immediately and while we're in here paying attention / testing it, seems best to move the next update point a year or two farther away ?

I think I miss understood the log at CircleCI:

```shell
[semantic-release]: node version >=16 || ^14.17 is required. Found v12.22.7.
```

I thought we needed specifically the `14.x` 😅

PR #1222, opened 2022-06-11
ci(node): setting publish version to 16

MateusAndrade:
# Overview

Updating publish-version job to 16

# Test Plan


mikehardy:
Ace!

PR #1231, opened 2022-06-22
fix(android): facebook stories

luisflorido:
# Overview

As said in [this issue](https://github.com/react-native-share/react-native-share/issues/1196), Facebook Stories is not working on Android, @steo-ml shows a fix [here](https://github.com/react-native-share/react-native-share/issues/1196#issuecomment-1120639001).
The image in Facebook Stories is using `ShareFile` constructor without the image type, some problem is happening to get the image type (probably a future fix on `ShareFile`), and the temporary fix is to use the constructor which you can send the type.

# Test Plan

Run Android app, and try to share Facebook Stories using a background image or sticker image

### DIFF
--- a/android/src/main/java/cl/json/social/FacebookStoriesShare.java
+++ b/android/src/main/java/cl/json/social/FacebookStoriesShare.java
@@ -98,14 +98,14 @@ public class FacebookStoriesShare extends SingleShareIntent {
                 backgroundFileName = options.getString("backgroundVideo");
             }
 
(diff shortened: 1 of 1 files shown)


MateusAndrade (approved):

steo-ml:
@luisflorido may I know if you're sharing the image via a base64 or a local file path? As I revisited this issue, I've found that the issue may actually be in the code where it reads base64 file, instead of `ShareFile` not specifying a file type. Plus, I find this hardcoding of file type into `ShareFile` to be a problem, as it breaks the functionality of sharing video backgrounds.

PR #1268, opened 2022-08-19
feat(ios): Share a link for Messenger on iOS

ken0nek:
# Overview

Add a feature to share URLs for Messenger on iOS

I want to port this implementation into this repo.

```js

Linking.openURL('fb-messenger://share?link=https://github.com/react-native-share/react-native-share');
```

ref: https://tools.applemediaservices.com/app/454638411?country=us

# Test Plan

### DIFF
--- a/ios/RNShare.m
+++ b/ios/RNShare.m
@@ -19,6 +19,7 @@
 #import "EmailShare.h"
 #import "TelegramShare.h"
 #import "ViberShare.h"
+#import "MessengerShare.h"
 #import "RNShareActivityItemSource.h"
 #import "RNShareUtils.h"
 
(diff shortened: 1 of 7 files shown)


mikehardy (approved):
This looks good to me (for a rebase merge even, as the two commits are separate and clean)
Definitely compiles locally but I didn't test it because I don't have an app that can do it

Any thoughts on adding it to the list of examples @ken0nek ? https://github.com/react-native-share/react-native-share/blob/main/example/App.js

ken0nek:
Thank you for the review!

I would like to add an example of this feature.

MateusAndrade:
> Thank you for the review!
> 
> I would like to add an example of this feature.

Nice! Do you think we can add the example separadetly?

MateusAndrade (approved):

ios/RNShare.m
```diff
@@ -84,6 +85,7 @@ - (NSDictionary *)constantsToExport
     @"INSTAGRAMSTORIES": @"instagramstories",
     @"TELEGRAM": @"telegram",
     @"EMAIL": @"email",
+    @"MESSENGER": @"messanger",
```
LeongXianJun, line 88: fyi, typo here causes failure to access messenger sharing if someone is using the enum exported in their shareSingle method

PR #1277, opened 2022-09-19
bug: show extra dialog on Android, when users share a video to instagram

KristineTrona:
# Overview
Fixes [this issue](https://github.com/react-native-share/react-native-share/issues/1276).

# Test Plan

1. Run the example app on Android, make sure you have Instagram installed on the device.
2. You should see a new button **"Share Video to IG"**
4. Pressing on **"Share Video to IG** opens up a dialog, where user can choose to share the video either directly, to feed, to reels or to stories.
5. As control also check that **"Share image to IG"** stills works as usual

<details>
<summary>See screen recording:</summary>

https://user-images.githubusercontent.com/39855778/192222553-43f32103-0b5a-46c5-b21f-ad7875a92845.MOV
</details>


mikehardy:
https://github.com/react-native-share/react-native-share/blob/ce90026ede1afdde44f0db1c18432cf326e3da08/android/src/main/java/cl/json/social/InstagramShare.java#L47-L54

This looks might suspicious, since you specify in your reproducer that it works with image/jpeg but not video/mp4

What are the assumptions in the code? There is clearly an assumption about image types vs non-image types - *why* does the code make that assumption? Was there some historical restriction on types that could be shared to instagram / instagram stories? Was it simply not a use case of the original PR author? Are there any restrictions like that now and does it work without them ?

Now that the example you've made cleanly reproduces it - I think you can open the `android/` folder from this repo directly in AndroidStudio (or VSCode...) and try making changes / rebuild and the dev cycle as you probe that chunk of code should be pretty speed https://react-native-share.github.io/react-native-share/docs/contributing#using-the-example-app

KristineTrona:
@mikehardy Ok, thank you for your input. I am not a Java developer, so might take me some time to get the right fix, but will try to look at it sometime later in the week. :)

mikehardy:
Oh I would definitely still leave this open! It's useful, and can easily serve in the future for anyone needing to demonstrate instagram stuff, hopefully you don't mind if I reopen it?

Separately - I will say, the java used in this module is just about the simplest java around, if you were ever thinking of just saying "what the heck" and hacking on some Java, *and* you were able to make a new PR here with reproduction (not something most can/will do), I'm about 99.999% sure you'll open up the java and it won't be bad to hack on at all. `System.err.println("REALLYOBVIOUSTAG: my thing here");` and `adb logcat |grep REALLYOBVIOUSTAG` are my goto tricks for quick and dirty java hacking like this.

I feel the same way the other direction mostly (I am terrible at Objective-C, really awful) but even I can hack on react-native objective-c code, because it's normally so simple.

KristineTrona:
Ok, thanks for the encouragement and tips, I will give it a try when I have some more time :)

KristineTrona:
@mikehardy I've adjusted the PR to fix the bug

android/src/main/java/cl/json/social/InstagramShare.java
```diff
@@ -61,18 +57,32 @@ protected void openInstagramUrlScheme(String url) {
...
         feedIntent.putExtra(Intent.EXTRA_STREAM, uri);
         feedIntent.setPackage(PACKAGE);
 
         Intent storiesIntent = new Intent("com.instagram.share.ADD_TO_STORY");
-        storiesIntent.setDataAndType(uri, "jpg");
+
+        if (isImage) {
+            storiesIntent.setDataAndType(uri, "jpg");
+        } else {
+            storiesIntent.setDataAndType(uri, "mp4");
```
MateusAndrade, line 83: This only accepts `mp4` files?

MateusAndrade (changes requested):
I think we might need to change our docs about share-single before merging this: https://react-native-share.github.io/react-native-share/docs/share-single 👀

android/src/main/java/cl/json/social/InstagramShare.java
KristineTrona: Hey @MateusAndrade I've updated the PR to allow also other type of files besides the `jpg` and `mp4` format

MateusAndrade: Ah, nice! Sorry for the delay, nice work! 😄

KristineTrona: Thanks! :)

mikehardy:
Nice!

Issue #1249, opened 2022-07-15
Share.open method doesn't work if user have more than one whatsapp installed in the device.


kv77724-dot:
### Steps to reproduce
1.Clone the whatsapp in you device.
2.Try to share anything using Share.open() method.
3.Click on whastapp logo. 

### Expected behaviour
Tell us what should happen
When trying to share using whatsapp it should open WhatsApp and let the user select the contact he want to share message to.

### Actual behaviour
Tell us what happens instead
When clicking on whatsapp logo, share Sheet closes automatically.

### Environment
System:
    OS: macOS 12.0.1
    CPU: (8) arm64 Apple M1
    Memory: 87.03 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.2 - /usr/local/bin/node
    Yarn: 1.22.17 - /opt/homebrew/bin/yarn
    npm: 7.21.1 - /opt/homebrew/bin/npm
    Watchman: 2021.11.15.00 - /opt/homebrew/bin/watchman
  Managers:
    CocoaPods: 1.11.2 - /opt/homebrew/bin/pod
  SDKs:
    iOS SDK:
      Platforms: DriverKit 21.0.1, iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0
    Android SDK:
      API Levels: 23, 28, 29, 30, 31
      Build Tools: 23.0.1, 28.0.3, 29.0.2, 29.0.3, 30.0.2, 30.0.3, 31.0.0
      System Images: android-24 | ARM 64 v8a, android-28 | Google APIs Intel x86 Atom, android-28 | Google Play Intel x86 Atom_64, android-29 | Google APIs ARM 64 v8a, android-30 | Google Play ARM 64 v8a, android-31 | Google APIs ARM 64 v8a, android-31 | Google APIs Intel x86 Atom_64
      Android NDK: Not Found
  IDEs:
    Android Studio: 2020.3 AI-203.7717.56.2031.7621141
    Xcode: 13.1/13A1030d - /usr/bin/xcodebuild
  Languages:
    Java: 18.0.1 - /usr/bin/javac
  npmPackages:
    @react-native-community/cli: Not Found
    react: 17.0.2 => 17.0.2 
    react-native: ^0.68.0 => 0.68.2 
    react-native-macos: Not Found
  npmGlobalPackages:
    *react-native*: Not Found
- **React Native platform + platform version**:Android 10,11, etc

[...]

PR #1287, opened 2022-10-08
WhatsApp share issue #1249

vijaychouhan-rails:
# Overview
shareSingle method doesn't work if the user has more than one WhatsApp(Cloned Whatsapp) installed on the device. Similar #1249

# Steps
1) Install WhatsApp in Android and Create a clone of the same, now you will have two whatsapp
2) Now Share.shareSingle(shareOptions);  with PDF file share
3) A window option will open and asking which whatsapp would you like to choose
4) Choose any Whatsapp and nothing will be inside conversation on Whatsapp

To fix above issue, this PR will solve that issue

### DIFF
--- a/android/src/main/java/cl/json/social/WhatsAppBusinessShare.java
+++ b/android/src/main/java/cl/json/social/WhatsAppBusinessShare.java
@@ -1,7 +1,7 @@
 package cl.json.social;
 
 import android.content.ActivityNotFoundException;
-import android.content.ComponentName;
+// import android.content.ComponentName;
 
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReadableMap;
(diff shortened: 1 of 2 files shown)


mikehardy:
In general, when using a system that has version control (that is: git, via github) leaving commented out code is a bad sign, it indicates you either do not have confidence in the code changes (so you can uncomment them later?) or you do not value clean code (because over time you will have increasingly irrelevant comments littered all over the code and no one will understand what the commented out code was doing or why it is still there.

So if you have a PR via git, prefer to just delete code that you no longer need

That said, you have some explanations - in comments - in the code but you have not actually explained *in the PR* or *in the commit message* why your changes have the important effect of fixing your problem. In fact, you have not actually described why the current code causes the problem.

Also it seems that this is the sort of thing affected by "clone" installs of whatsapp, but the fix will affect people that just have a single whatsapp install, so the test plan should:

- describe what a cloned whatsapp install is and how to set up a test scenario with cloned whatsapp
- also include test on a device that just has a single whatsapp install so we know it does not affect the regular use case

PR #1292, opened 2022-10-14
Document that error is thrown when sheet is closed

princefishthrower:
# Overview

As the title states, this small doc change would have saved me from the issue I raised in https://github.com/react-native-share/react-native-share/issues/1291

Totally flexible about the wording or arrangement, I think it's just important to mention somewhere in the docs that closing the sheet throws an error.

# Test Plan

No code changes, just a documentation improvement :)

### DIFF
--- a/website/docs/share-single.mdx
+++ b/website/docs/share-single.mdx
@@ -41,6 +41,16 @@ Or with `async/await`:
   }
 ```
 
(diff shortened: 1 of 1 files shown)


MateusAndrade (approved):
🚀

mikehardy:
Nice one! Thanks

Issue #1304, opened 2022-10-29
Developer wanted to enhance instagram / facebook stories share with app ID parameter


zerakacordo:
### Steps to reproduce
1. Share to Instagram or Facebook with Stories now requires Facebook App ID in their request as of October 2022.
2. Read at https://developers.facebook.com/docs/instagram/sharing-to-stories

### Expected behaviour
Should allow a parameter for Facebook APP ID.

### Actual behaviour
-

### Environment
-

### react-native-share
7.9.1

mikehardy:
Happy to merge any + all reasonable PRs!

mikehardy:
https://developers.facebook.com/blog/post/2022/10/10/introducing-important-update-to-Instagram-sharing-to-stories/

> All developers who have integrated previously without sending a Facebook App ID will need to make this change by no later than November 30, 2022, to avoid disruption to their integration with our platform. After November 30, 2022, users of your app will lose access to share your content to Instagram Stories if you have not implemented this change.

https://developers.facebook.com/docs/instagram/sharing-to-stories/

ZerakPalani:
Thank you for the quick response! Thought I would share it here. If nobody gets it fixed I might try to create a PR to it, I am just not sure exactly how it works yet (I will look into all of the features I need for my app and try to create PRs so that everybody can have these).

PR #1305, opened 2022-10-31
feat(android, iOS)!: require appId for instagram stories to meet new Meta requirement

rickafds:
# Overview

Implemented #1304

# Test Plan

https://developers.facebook.com/blog/post/2022/10/10/introducing-important-update-to-Instagram-sharing-to-stories/
https://developers.facebook.com/docs/instagram/sharing-to-stories/

### DIFF
--- a/android/src/main/java/cl/json/social/InstagramStoriesShare.java
+++ b/android/src/main/java/cl/json/social/InstagramStoriesShare.java
@@ -49,10 +49,6 @@ public class InstagramStoriesShare extends SingleShareIntent {
     }
 
     private void shareStory(ReadableMap options) {
-        if (!this.hasValidKey("appId", options)) {
-            throw new IllegalArgumentException("appId was not provided.");
-        }
-
         if (!this.hasValidKey("backgroundImage", options) && !this.hasValidKey("backgroundVideo", options)
                 && !this.hasValidKey("stickerImage", options)) {
             throw new IllegalArgumentException("Invalid background or sticker assets provided.");
--- a/android/src/main/java/cl/json/social/InstagramStoriesShare.java
+++ b/android/src/main/java/cl/json/social/InstagramStoriesShare.java
@@ -49,6 +49,10 @@ public class InstagramStoriesShare extends SingleShareIntent {
     }
 
     private void shareStory(ReadableMap options) {
+        if (!this.hasValidKey("appId", options)) {
+            throw new IllegalArgumentException("appId was not provided.");
+        }
+
         if (!this.hasValidKey("backgroundImage", options) && !this.hasValidKey("backgroundVideo", options)
                 && !this.hasValidKey("stickerImage", options)) {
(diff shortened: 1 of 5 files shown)


android/src/main/java/cl/json/social/InstagramStoriesShare.java
```diff
@@ -49,6 +49,10 @@ protected String getPlayStoreLink() {
     }
 
     private void shareStory(ReadableMap options) {
+        if (!this.hasValidKey("appId", options)) {
+            throw new IllegalArgumentException("appId was not provided.");
+        }
+
```
mikehardy, line 55: I like the idea of failing fast for developers, and I see you are error checking here (for missing appId), but on the iOS side, not checking

I wonder if you could check in the javascript code and fail there with an error message, but slightly more descriptive, for example "Instagram Story share requires an appId based on Meta policy" or similar?

That way (in javascript) it would fail on both platforms, with a helpful message

mikehardy (approved):
I like it! Fail fast. We'll have to do a major release on this since it will break existing code, but that is totally justified + okay to do

mikehardy:
@MateusAndrade this looks good to me - I put a `!` in the PR title so it should :crossed_fingers: trigger a major release, with the PR message indicating what exactly the break is. I read the related docs on the issue when it first came through, this all seems :100:  to me

@rickafds --> :trophy: ! Thank you for the contribution

android/src/main/java/cl/json/social/InstagramStoriesShare.java
rickafds: @mikehardy thanks for the support

ZerakPalani:
@rickafds Thank you for contributing!

MateusAndrade:
> @MateusAndrade this looks good to me - I put a `!` in the PR title so it should 🤞 trigger a major release, with the PR message indicating what exactly the break is. I read the related docs on the issue when it first came through, this all seems 💯 to me
> 
> @rickafds --> 🏆 ! Thank you for the contribution

I forgot to merge this with the breaking change pattern, so I did an `ammend` on the commit and now I'm a co-author with @rickafds. 😅 

Btw, thanks for the amazing work @rickafds! 🚀  🚀  🚀

src/types.ts
```diff
@@ -74,6 +74,7 @@ interface BaseSocialStoriesShareSingleOptions extends Omit<BaseShareSingleOption
 
 export interface InstagramStoriesShareSingleOptions extends BaseSocialStoriesShareSingleOptions {
   social: Social.InstagramStories;
+  appId: string;
```
BrodaNoel, line 77: Can somebody add more info here? What appId? Should this be the Facebook App Id? Or the AppStore app id?

alis0nc: According to [Facebook's docs](https://developers.facebook.com/docs/instagram/sharing-to-stories/) on sharing to Instagram stories, the id is the Facebook app id.

Issue #760, opened 2020-04-12
Unable to share image+text IOS


sashaskuku:
### Steps to reproduce

code looks like that:

await RNShare.open({
	message: singleItem.name,
	url: imageLink
		});

### Expected behaviour
Share image AND text

### Actual behaviour
This code sends only text. If I remove message at all - it sends image only. 
I tried also to use base64 to send Image with no success.
### Environment
- **React Native version**:"0.61.2",
- **React Native platform + platform version**: iOS 9.0

### react-native-share
**Version**: 2.0.0

meetyourdev:
i am facing this issue as well

MateusAndrade:
This happens only on iOS on a specific app? I remember having a similar issue on WhatsApp. I ended removing the message param on iOS. 😢 

#474

hungneox:
I had similar issue when sharing pdf + text to Slack. Does anyone has the same issue?

FrederickEngelhardt:
Same issue experienced with messages app. It works for me with emails.

MateusAndrade:
> Same issue experienced with messages app. It works for me with emails.

Which platforms are you facing this? Probably this is a problem with the target app. Unfortunately, I faced this on a project and decided to share only the image(it was receipt screen).

joedivittorio:
Seeing a similar issue on iOS. If you add an image AND text, the iOS messages app only uses the image. Able to send only text or only an image but not both on the messages app. Seeing the same behavior on other apps as well like WhatsApp and Slack.

For the mail app, the subject, text, and an image all work as expected with the same parameters.

This is all using the method `RNShare.open`

agnieszkabugla:
I have the same issue on iOS when I want to share an image + message:
- via whatsapp -> only message is shared
- via messanger -> only image is shared
- via mail -> works as supposed, message and image are shared

```

shareSingleImage = async () => {
    const url = 'https://awesome.contents.com/';
    const title = 'Awesome Contents';
    const message = 'Please check this out.';

    const shareOptions = {
      title: 'Share file',
      url: `data:image/png;base64,${this.state.base64}`,
      subject: title,
      message: `${message} ${url}`,
      failOnCancel: false,
    };

    try {
       await Share.open(shareOptions);
    } catch (error) {
      console.log('Error =>', error);
    }
  };
```

MateusAndrade:
[...]

Issue #280, opened 2018-06-18
Image Sharing for WhatsApp is not working


iSaadSalman:
So I'm saving a screenshot locally and passing the URI to Share options. Its working fine for every platform except Whats App.
 
`let shareOptions = {
      title: "Share",
      url: this.state.cameraRollUri,
      type: 'image/png',
      message: "Here it is",
      subject: "Share Link", //  for email,
    };

  
    Share.open(shareOptions)`

Just message appears but no image

jgcmarins:
@iSaadSalman application is crashing?
Do you have a stack trace?

iSaadSalman:
No it’s not crashing. Just the image is not being shared in WhatsApp. Check the description I wrote in op.

jgcmarins:
Is it happening on both platforms?

iSaadSalman:
@jgcmarins Just testing on iOS right now. It's happening there.

iSaadSalman:
@jgcmarins Do you have any idea what could be the problem?

gaganjot-singh01:
Same problem

jgcmarins:
@iSaadSalman can you try master?

iSaadSalman:
@jgcmarins I tried by overwriting the module in node_modules since I'm not able to update via npm.

I'm receiving this build error in Xcode:
` Undefined symbols for architecture arm64:
  "_OBJC_CLASS_$_InstagramShare", referenced from:
      objc-class-ref in libRNShare.a(RNShare.o)
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)`

iSaadSalman:
@jgcmarins any update?

jgcmarins:
@iSaadSalman I could not take a deeper look.
But if you want to help us, PRs are welcome.

jgcmarins:
> @jgcmarins I tried by overwriting the module in node_modules since I'm not able to update via npm.
> 
> I'm receiving this build error in Xcode:
> Undefined symbols for architecture arm64: "_OBJC_CLASS_$_InstagramShare", referenced from: objc-class-ref in libRNShare.a(RNShare.o) ld: symbol(s) not found for architecture arm64 clang: error: linker command failed with exit code 1 (use -v to see invocation)

Your link is wrong here.
I recommend to remove all node_modules, install lib again, following [README](https://github.com/react-native-community/react-native-share/blob/master/README.md) and also link again.

[...]

PR #1332, opened 2022-12-19
Allow custom activity types to be specified inside activityItemSources

hieu9102002:
# Overview

A few applications' api do not allow for sending of both messages and url (E.g.: #280 #760 WhatsApp, Slack). When this happens, the app api randomly chooses either the message or the url to be shared to the application (for WhatsApp, the message is always prioritized over the url). We would like to specify which of the two to send for these type of application when calling Share.open using activityItemSources.

# Test Plan

Inside where I am calling Share.open, I can now specify the bundle to specifically tell share options what to send and what not to, instead of just using the default ActivityTypes.
```ts
const url = payload.url ?? '';
const message = payload.message ?? '';
const options: ShareOptions = Platform.select({
	ios: {
		activityItemSources: [
			{
				placeholderItem: { type: 'url', content: url },
				item: { default: { type: 'url', content: url } },
			},
			{
				placeholderItem: { type: 'text', content: message },
				item: {
					default: { type: 'text', content: message },
					'net.whatsapp.WhatsApp.ShareExtension': null,
				},
			},
		],
	},
	default: {
		...payload,
	},
});
```

To verify the change works, try sharing using WhatsApp, now only the url is sent, the message is not. For other app shares, both the url and the message is still sent.

### DIFF
--- a/ios/RNShareActivityItemSource.m
+++ b/ios/RNShareActivityItemSource.m
@@ -197,7 +197,7 @@
         }
     }
     
-    return nil;
+    return activityType;
 }
 
 + (nullable id)objectForActivityType:(UIActivityType)activityType inDictionary:(nonnull NSDictionary *)dictionary {
--- a/src/types.ts
+++ b/src/types.ts
@@ -42,10 +42,10 @@ export interface ActivityItem {
 
 export interface ActivityItemSource {
   placeholderItem: ActivityItem;
-  item: { [key in ActivityType]?: ActivityItem | null | undefined };
-  subject?: { [key in ActivityType]?: string };
-  dataTypeIdentifier?: { [key in ActivityType]?: string };
-  thumbnailImage?: { [key in ActivityType]?: string };
+  item: { [key in ActivityTypeItemSource]?: ActivityItem | null | undefined };
+  subject?: { [key in ActivityTypeItemSource]?: string };
+  dataTypeIdentifier?: { [key in ActivityTypeItemSource]?: string };
+  thumbnailImage?: { [key in ActivityTypeItemSource]?: string };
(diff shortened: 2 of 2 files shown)


mikehardy (approved):
Hmm - not sure *exactly* how it does that, but it's a non-breaking change that appears to be mostly just providing more information as a return value, can't see how that will cause active harm :shrug:

mikehardy:
For the avoidance of doubt: this seems to  be iOS-only? You don't mention that but it appears to be the only affected platform.

hieu9102002:
> Hmm - not sure exactly how it does that

At the moment, if you specify an activity that is not in the list of default UI Activity items, say `net.whatsapp.WhatsApp.ShareExtension` inside my code snippet for example, it will be ignored. By making the change from returning nil if the users' provided activity items is not in the list of defaults to returning itself, it allows the UI activity view controller to attempt to find the share bundle the users' requested and make specific changes to that bundle. In this example, for every application but the one with the bundle name `net.whatsapp.WhatsApp.ShareExtension`, both the url and the message will be shared, but for WhatsApp sharing, it will only send the url and not the message, similarly to the example in [the docs](https://react-native-share.github.io/react-native-share/docs/share-open/#activityitemsources-ios-only), but instead of working for the messages app, it works for WhatsApp.

The system already have checks against bad input, so in case users input a share bundle name that does not exist, it will treat that as basically nil, so this change is non-breaking.

> For the avoidance of doubt: this seems to be iOS-only? You don't mention that but it appears to be the only affected platform.

Yes. UI Activity items are ios-only.

@mikehardy Hopefully my answer was detailed enough, please let me know if anything is still unclear.

PR #1336, opened 2022-12-24
Refactor UI components to function components

whiteHatCip:
# Overview
This MR aims at providing a new implementation of the UI components that makes use of function components, moving away from the class components that were already in place.

# Test Plan

### DIFF
--- a/src/components/Overlay.tsx
+++ b/src/components/Overlay.tsx
@@ -23,41 +23,36 @@ export interface OverlayProps {
   visible: boolean;
 }
 
-interface State {
-  fadeAnim: Animated.Value;
-  overlayStyle: StyleProp<ViewStyle>;
-}
+const Overlay: React.FC<React.PropsWithChildren<OverlayProps>> = ({
+  visible,
+  children,
+}) => {
+  const [fadeAnim] = React.useState(new Animated.Value(0));
+  const [overlayStyle, setOverlayStyle] = React.useState<StyleProp<ViewStyle>>(styles.emptyOverlay);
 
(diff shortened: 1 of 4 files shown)


src/components/Overlay.tsx
```diff
@@ -23,41 +23,28 @@ export interface OverlayProps {
   visible: boolean;
 }
 
-interface State {
-  fadeAnim: Animated.Value;
-  overlayStyle: StyleProp<ViewStyle>;
-}
+const Overlay: React.FC<React.PropsWithChildren<OverlayProps>> = ({ visible, children }) => {
+  const [fadeAnim] = React.useState(new Animated.Value(0));
+  const [overlayStyle, setOverlayStyle] = React.useState<StyleProp<ViewStyle>>(styles.emptyOverlay);
```
MateusAndrade, line 28: What about importing the `useState` separately from React?

whiteHatCip: > What about importing the `useState` separately from React? 

We can do that, of course. But I don't see that much difference imho. Will make the change when I'm back from holidays tho😉

MateusAndrade (approved):

whiteHatCip:
The docs page should remove the "Refactor to function components" element from the todo list at this link:
https://react-native-share.github.io/react-native-share/docs/ui-components

MateusAndrade:
> The docs page should remove the "Refactor to function components" element from the todo list at this link: https://react-native-share.github.io/react-native-share/docs/ui-components

Done: https://github.com/react-native-share/react-native-share/commit/fd94ea54f74dd17d054b760256023ece351f9296! 🚀

Issue #1352, opened 2023-01-19
Having issues sharing to snapchat on android 13


Maxarreaga:
### Steps to reproduce
1. Have snapchat installed 
2. Try to share to snapchat from a device running android 13
3. Get error: 
Error: {"error":"Unable to find explicit activity class {com.snapchat.android/com.snapchat.android.LandingPageActivity}; have you declared this activity in your AndroidManifest.xml, or does your intent not match its declared <intent-filter>?"}

### Expected behaviour
Picture should share to snapchat 

### Actual behaviour
Error is returned and  picture is not shared to snapchat

### Environment
- **React Native version**:  0.70.5
- **React Native platform + platform version**: Android 13
- **Typescript version** (if using typescript): 4.5.5

### react-native-share
**Version**: 8.1.0

Also notably changing the CLASS variable in SnapChatShare.java  from "com.snapchat.android.LandingPageActivity" to "com.snap.mushroom.MainActivity" fixes the problem, or if the getComponentClass function is replaced all together (making the returned class null by default)

Would appreciate some repros, thanks!

EDIT: it seems related to [the following change in Android 13](https://developer.android.com/about/versions/13/behavior-changes-all#:~:text=Intent%20filters%20block%20non-matching%20intents%20When%20your%20app%2cin%20the%20receiving%20app.%20Non-matching%20intents%20are%20blocked.)

Security
Intent filters block non-matching intents
When your app sends an intent to an exported component of another app that targets Android 13 or higher, that intent is delivered if and only if it matches an <intent-filter> element in the receiving app. Non-matching intents are blocked.

Similarly, if your app upgrades to Android 13 or higher, any intent sent from another app is delivered to an exported component of your app if and only if that intent matches a declared <intent-filter> element in your app.

I'm guessing that the original class  mentioned was/isn't actually in the snapchat activity filter?

bibhuti9:
Try this code into your androidmanifset

 ```
 <queries>
      <package android:name="com.google.android.youtube" />
      <package android:name="com.facebook.katana" />
      <package android:name="com.facebook.orca" />
      <package android:name="com.facebook.mlite" />
      <package android:name="com.facebook.lite" />
      <package android:name="org.telegram.messenger" />
      <package android:name="com.whatsapp" />
      <package android:name="com.instagram.android" />
    </queries>
```


<manifest xmlns:android="http://schemas.android.com/apk/res/android"
  package="com.technobot.quotesdaily">


     <queries>
[...]

PR #1353, opened 2023-01-24
fixing snapchat on android 13

Maxarreaga:
# Overview
Android 13 seems to have made changes to the intent filters and what it allows through. This PR is a fix to the issue mentioned in #1352 that I ran into previously. Since this intent is not documented anywhere, I had to find the string by seeing what other apps called in their requests to snapchat. The excerpt is below. And the change is confirmed to work on both Android 12 and 13.

> Security
> Intent filters block non-matching intents
> When your app sends an intent to an exported component of another app that targets Android 13 or higher, that intent is delivered if and only if it matches an element in the receiving app. Non-matching intents are blocked.
> 
> Similarly, if your app upgrades to Android 13 or higher, any intent sent from another app is delivered to an exported component of your app if and only if that intent matches a declared element in your app.

# Test Plan
Tested the change using an app that react-native-share to share to snapchat and verified it worked on Android 12 and 13


MateusAndrade:
Does it work with Android version below 11?

Maxarreaga:
I don’t have a device to test it but I don’t see why it wouldn’t work since the Snapchat intent was introduced in 2019ish

MateusAndrade (approved):

PR #1383, opened 2023-04-13
fix(android): instagram direct share plain text

TimP4w:
# Overview
Hi there,
We are using your library at our company, however we had problems sharing direct messages on instagram with android.
The current behaviour is that nothing happens if a url is not provided, and when provided it says that the file type is not supported. 
This patch tries to fix that by passing `text/*` as type in the options.

# Test Plan
In the example app simply try to send an instagram direct message.

Cheers


MateusAndrade (approved):

MateusAndrade:
Nice, thanks @TimP4w !

jeffreynerona:
When using Share.open instead of Share.share, instagram is not in the options

PR #1402, opened 2023-06-19
@thomasfetiveau/newarch

zabojad:
# Overview
This PR adds support for both new and old architecture (Turbo modules).

# Test Plan
I've tested it successfully on React native 0.68 and React native 0.71 with both old and new architecture.


ios/RNShare.mm
```diff
@@ -300,17 +307,57 @@ - (NSDictionary *)constantsToExport
...
     }
 }
 
+# pragma mark - New Architecture
+
+#if RCT_NEW_ARCH_ENABLED
+
+// - (facebook::react::ModuleConstants<JS::NativePermissionsModule::Constants::Builder>)getConstants {
+//   return [self constantsToExport];
+// }
```
MateusAndrade, line 353: Can we remove this?

ios/RNShare.mm
```diff
@@ -300,17 +307,57 @@ - (NSDictionary *)constantsToExport
...
         NSError *error = [NSError errorWithDomain:@"CANCELLED" code: 500 userInfo:@{NSLocalizedDescriptionKey:@"PICKER_WAS_CANCELLED"}];
-        rejectBlock(error);
+        rejectBlock(@"CANCELLED",@"CANCELLED",error);
     }
 }
 
 - (void)documentPicker:(UIDocumentPickerViewController *)controller didPickDocumentsAtURLs:(NSArray<NSURL *> *)urls {
     if (resolveBlock) {
-        resolveBlock(@[@(YES), @"com.apple.DocumentsApp"]);
+        // resolveBlock(@[@(YES), @"com.apple.DocumentsApp"]);
```
MateusAndrade, line 339:
```suggestion
```

zabojad:
Hey @MateusAndrade ! It should be OK now, please have a look :).

MateusAndrade:
> Details

Niceee, the JS steps are working, and Android as well! But we still have an issue with iOS: https://app.circleci.com/pipelines/github/react-native-share/react-native-share/1073/workflows/c0dfa209-c2a6-41fd-802b-1fb23c5b0293/jobs/4134.

Besides that it looks great, let me know if you need any help with that.

zabojad:
@MateusAndrade The example seems to be using an old version of RN that doesn't know the `install_modules_dependencies`  podfile method...

How can that be updated?

zabojad:
@MateusAndrade if you could give a hand on this... I do not get why the ios build of the example is not passing...

EstebanFuentealba:
Hello, thank you for contributing to the project. I'm not entirely sure as I haven't worked with React Native for a while, but the documentation outlines certain things that need to be implemented for the new architecture on iOS. You can find more information at https://reactnative.dev/docs/new-architecture-library-ios.


1. Updating your Podspec for the New Architecture
The New Architecture makes use of CocoaPods.

Add Folly and Other Dependencies
The New Architecture requires some specific dependencies to work properly. You can set up your podspec to automatically install the required dependencies by modifying the .podspec file. In your Pod::Spec.new block, add the following line:

```diff
Pod::Spec.new do |s|
  # ...
+  install_modules_dependencies(s)
  # ...
end
```

At this [link](https://github.com/facebook/react-native/blob/82e9c6ad611f1fb816de056ff031716f8cb24b4e/scripts/react_native_pods.rb#L139-L144), you can find the documentation of the install_modules_dependencies function.

If you need to explicitly know which folly_flags React Native is using, you can query them using the [folly_flag](https://github.com/facebook/react-native/blob/82e9c6ad611f1fb816de056ff031716f8cb24b4e/scripts/react_native_pods.rb#L135) function.

EstebanFuentealba:
Would it be a good idea to create a sample project from scratch?

zabojad:
@EstebanFuentealba the problem with `install_modules_dependencies` is that it works only on RN 0.71+ project... I've fixed this issue already. I now have other issues with circleci that are not related with that...

zabojad:
@MateusAndrade @EstebanFuentealba  OK ! It seems ok now. Please have a look...

MateusAndrade:
Niceeeeee! Thanks a lot @zabojad , @EstebanFuentealba ! 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀 🚀

MateusAndrade:
> @MateusAndrade we should add a note in the README that new arch is also supported now... Thanks for having merged the PR!

Ah yes! I will do so, I needed do run a amend with your commit after merging it. 😅

PR #1408, opened 2023-06-30
fix: add documentation for the newly required appId for ig stories

ameliagoodman:
# Overview
since jan 2023, facebook requires appId for sharing to stories. adding that to the documentation so people know they need to add it

# Test Plan
no need , just updated docs. the code for appId already existed


MateusAndrade (approved):
Thanks! 🚀

PR #1409, opened 2023-07-01
Cleanup promise nesting and make code a bit more maintainable

robinsoncol:
# Overview
I took a jab at removing the nested promise logic in `index.tsx`, clarifying logic where possible and making the code a bit more maintainable. I tried my best to preserve the order of things so there shouldn't be any breaking changes.

I marked a few areas of concern (comments that start with `Concern:`).  Those should get removed/resolved before we merge this PR (if we ever get to that point).

# Test Plan
If someone could test these changes, that would be great!


robinsoncol:
Managed to get the Android example to run. Found a minor bug, but everything else is looking good. 

Now, I just need to get the iOS example to run to further test the changes.

MateusAndrade:
> Managed to get the Android example to run.
> 
> Now, I just need to get the iOS example to run to further test the changes.

Nice, let me know if you need any support with that, this PR looks 🔥 🔥 🔥 🔥 .

MateusAndrade (approved):

robinsoncol:
Sweet! Let me know if anything is off and I'll look into it

PR #1438, opened 2023-08-03
Fixed getDataColumn not handling exception thrown by cursor.getColumn…

DylanJCraggs:
…IndexOrThrow

# Overview

# Test Plan

### DIFF
--- a/android/src/main/java/cl/json/RNSharePathUtil.java
+++ b/android/src/main/java/cl/json/RNSharePathUtil.java
@@ -161,6 +161,9 @@ public class RNSharePathUtil {
                 final int index = cursor.getColumnIndexOrThrow(column);
                 return cursor.getString(index);
             }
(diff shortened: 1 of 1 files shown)


DylanJCraggs:
An exception was being thrown attempting to open .msg file in Android which prevented the bottom sheet from opening.

This fix catches the exception and now displays the app sheet as expected.

MateusAndrade:
Could you share the details about this exception and how to reproduce it?

DylanJCraggs:
> Could you share the details about this exception and how to reproduce it?

@MateusAndrade 

Using Android 13, call Share.Open like below for a .msg file.

Share.open({
    url: fileUri, // make this a .msg file
    failOnCancel: false,
    filename: fileName,
    showAppsToView: true
});

MateusAndrade (approved):

PR #1569, opened 2024-09-02
fix(ios): import from codegen with RN 0.75+

huextrat:
# Overview

fixes: https://github.com/react-native-share/react-native-share/issues/1568

# Test Plan

### DIFF
--- a/ios/RNShare.h
+++ b/ios/RNShare.h
@@ -9,7 +9,7 @@
 #endif
 
 #if RCT_NEW_ARCH_ENABLED
-#import <RNShareSpec/RNShareSpec.h>
+#import <RNShareSpec.h>
 #endif
 
 @interface RNShare : NSObject <RCTBridgeModule, UIDocumentPickerDelegate>
--- a/ios/RNShare.mm
+++ b/ios/RNShare.mm
@@ -26,7 +26,7 @@
 #import "RNShareUtils.h"
 
 #if RCT_NEW_ARCH_ENABLED
-#import <RNShareSpec/RNShareSpec.h>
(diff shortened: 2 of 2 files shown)


MateusAndrade (approved):

huextrat:
@MateusAndrade I haven't tested it on RN 0.74 with new arch but this change may break the old RN versions on the new arch

MateusAndrade:
> @MateusAndrade I haven't tested it on RN 0.74 with new arch but this change may break the old RN versions on the new arch

I see, we can release a new major release. It looks like the pipeline for iOS is not working. 🤔

huextrat:
I don't know if it's related, it fails as soon as the job starts

`ios_build` job seems to also be on failure on previous PRs. Something related to CircleCI config ?

MateusAndrade:
> I don't know if it's related, it fails as soon as the job starts
> 
> `ios_build` job seems to also be on failure on previous PRs. Something related to CircleCI config ?

I see, it's related to the resource class used. I can fix this with another PR.

efstathiosntonas:
hey folks, I believe it needs double quotes instead of brackets, at least this is what Xcode is complaining about.

ie. `#import <RNShareSpec.h>` -> `#import "RNShareSpec.h"`

Screenshot 2024-09-02 at 15 20 33

huextrat:
Indeed! https://github.com/react-native-share/react-native-share/pull/1570

benji-jo:
Appeared again in version 0.76.3

saadlatif75 (commented):
yes this commit woiking fine...

PR #1588, opened 2024-10-15
Updated deprecated  [[UIApplication sharedApplication] openURL]] for iOS 18 & Xcode 16

vivekjoshi007:
**Bug** :-  WhatsApp share not working on iOS 18 while using Xcode 16.

**Reason** :-  [[UIApplication sharedApplication] openURL]] does not work on iOS 18 if built via Xcode 16

**Fix** :- Replacing the above with [[UIApplication sharedApplication] openURL: options: completionHandler:] fixed the issue

# Overview

# Test Plan

### DIFF
--- a/ios/FacebookStories.m
+++ b/ios/FacebookStories.m
@@ -87,7 +87,7 @@ RCT_EXPORT_MODULE();
     // Cannot open facebook
     NSString *stringURL = @"https://itunes.apple.com/app/facebook/id284882215";
     NSURL *url = [NSURL URLWithString:stringURL];
-    [[UIApplication sharedApplication] openURL:url];
+    [[UIApplication sharedApplication] openURL:url options:@{} completionHandler:nil];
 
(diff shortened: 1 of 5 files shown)


MateusAndrade (approved):

mikehardy (commented):
I'm always nervous when I see changes like this with no positive assertions on compatibility based on the source docs so I checked:

https://developer.apple.com/documentation/uikit/uiapplication/1622961-openurl?language=objc

indeed "calling this method has no effect" because of the deprecation I suppose, final phase before removal but really that's evil to leave it in like that, better to have a build break then a behavior break. But anyway

https://developer.apple.com/documentation/uikit/uiapplication/1648685-openurl?language=objc 

that's the new method and compat is listed as:

```
iOS 10.0+
iPadOS 10.0+
Mac Catalyst 13.1+
tvOS 10.0+
visionOS 1.0+
```

So should be fine ! Excellent

PR #1599, opened 2024-11-15
New Feature: add base64 image support under linkMetadata for IOS for custom sharing icon

hazrid93:
# Overview

Enhancement to library [react-native-share](https://github.com/react-native-share/react-native-share) to load share icon image from base64.

require usage of changeOption to this format:

```
const shareOptions = Platform.select({
                ios: {
                    activityItemSources: [
                        {
                            placeholderItem: {
                                type: 'text',
                                content: '...'
                            },
                            item: {
                                type: 'text',
                                content: '...'
                            },
                            linkMetadata: {
                                title: 'Title M2U Receipt',
                                base64Icon: `${yourLogo}` <-- Note: pass base64 image data here, image can be converted to base64 and define directly in file. Instead of live conversion.
                            }
                        },
                    ],
                },
                default: {
                    title: 'Receipt',
                },
            });
```

67990724-f8a2-435f-8537-a033e1f517bd

### DIFF
--- a/ios/RNShareActivityItemSource.m
+++ b/ios/RNShareActivityItemSource.m
@@ -102,6 +102,39 @@
     return nil;
 }
 
++ (UIImage *)adjustedForShareSheetPreviewIconProvider:(UIImage *)image {
+    UIColor *replaceTransparencyWithColor = [UIColor whiteColor]; // Change as required
+    CGFloat minimumSize = 40.0;  // Points
+
+    UIGraphicsImageRendererFormat *format = [[UIGraphicsImageRendererFormat alloc] init];
+    format.opaque = YES;
+    format.scale = image.scale;
+
+    CGFloat imageWidth = image.size.width;
+    CGFloat imageHeight = image.size.height;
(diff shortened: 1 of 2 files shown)


mikehardy (commented):
That looks pretty neat - image on PR appears to show it working well, only thing I can think of is that the specific LinkMetadata docs need a corresponding update and it might be good to also update the area in the document above LinkMetadata section where there is example usage and show how to use it ? (ideally showing a "you might try this to load an icon into base64" chunk of code so people have a full idea how to do it from icon image file to setting the property

https://github.com/react-native-share/react-native-share/blob/main/website/docs/share-open.mdx#linkmetadata

hazrid93:
@mikehardy sure

mikehardy (approved):
looks great to me, nice feature

mikehardy:
I haven't worked in this repo enough in long enough that I don't feel right merging without @MateusAndrade taking a look - sorry for the ping Mateus but it's just out of respect 😅  I don't want to break anything when you've been doing all the good work around here 😄

hazrid93:
Thanks @mikehardy !

PR #1612, opened 2024-12-10
fix(ios): handle local file on InstagramStories

huextrat:
# Overview

Since this PR https://github.com/react-native-share/react-native-share/pull/1598 (v11.1.0) sharing a local file to share it in instagram story no longer works.

At the same time, sharing via a remote URL doesn't seem to work either.

My suggestion is to keep @roli5005 code because it seems to work for him. But if we're sharing a local file with the scheme `file` then we're using a much simpler logic.

# Test Plan

Our pre-11.1.0 implementation works again thanks to this change


mikehardy:
This seems important as the referenced PR did seem valid (I remember looking at it) but clearly caused a regression
@MateusAndrade - have you seen this? Not sure if it's the best approach but coulud work

MateusAndrade:
> This seems important as the referenced PR did seem valid (I remember looking at it) but clearly caused a regression @MateusAndrade - have you seen this? Not sure if it's the best approach but coulud work

I see. I will update the branch, and we can proceed if the pipeline works for it 🤞

MateusAndrade (approved):

PR #1616, opened 2024-12-12
fix(iOS): update deprecated URL opening method for Telegram share

fekajin:
This change is aiming to change deprecated version of openURL method to new usage: `openURL:shareURL options:@{} completionHandler:^(BOOL success) {}`

# Test Plan

 versions: 
`"react-native": "0.75.4",
"react-native-share": "^12.0.3",
`
ios: 
`Xcode: 12.0,
Device IOS Version: 18.1.1
`

this is my implementation: 

`const shareOptions = {
          social: Share.Social.TELEGRAM,
          message: 'message',
          title: 'Share'
        };
`
`await Share.shareSingle(shareOptions);`

without any change, the singleShare is giving this log result for telegram options in Xcode but nothing happens in the ui:
  `telegram`
   `TRY OPEN telegram
BUG IN CLIENT OF UIKIT: The caller of UIApplication.openURL(_:) needs to migrate to the non-deprecated UIApplication.open(_:options:completionHandler:). Force returning false (NO).`

after the deprecated URL change, this is the response i got: 
`telegram
TRY OPEN telegram`

And telegram app shows up with share screen after this logs.


ios/TelegramShare.m
```diff
@@ -27,7 +27,7 @@ - (void)shareSingle:(NSDictionary *)options
     NSURL * shareURL = [NSURL URLWithString:urlTelegram];
     
     if ([[UIApplication sharedApplication] canOpenURL: shareURL]) {
-        [[UIApplication sharedApplication] openURL: shareURL];
+       [[UIApplication sharedApplication] openURL:shareURL options:@{} completionHandler:^(BOOL success) {}];
```
mikehardy, line 30: shouldn't the `resolve(@[@true, @""]);` be in the completion handler ? so that the promise resolves when the share is complete

fekajin: Thanks for the feedback, you're right. First i did it like that and then i checked how its handled for other share options such as Instagram in **InstagramShare.m** : 
`if ([[UIApplication sharedApplication] canOpenURL: shareURL]) {
        [[UIApplication sharedApplication] openURL:shareURL options:@{} completionHandler:nil];
        resolve(@[@true, @""]);
    }` 

I didn't wanted to implement different logic from the other sharing options, but i can change it like this if its okay:
```
if ([[UIApplication sharedApplication] canOpenURL:shareURL]) {
        [[UIApplication sharedApplication] openURL:shareURL options:@{} completionHandler:^(BOOL success) {
            if (success) {
                resolve(@[@true, @"Successfully opened Telegram"]);
            } else {
                NSString *errorMessage = @"Failed to open Telegram for an unknown reason";
                NSError *error = [NSError errorWithDomain:@"com.rnshare" code:2 
                                                userInfo:@{NSLocalizedDescriptionKey: errorMessage}];
                reject(@"Open Error", errorMessage, error);
            }
        }];
    } 
```

But i do not know in which cases completionHandler returns false with the successfull canOpenURL condition. (Maybe this case happens if the Telegram crashes because of some reason?)

mikehardy: your proposed new code *seems* more correct. I'm not sure why the other ones aren't doing that but if they are not, *and* your code works well for you (testing expected success and failure modes) *then* it seems like your code should be the style we use here and likely everywhere...

PR #1618, opened 2024-12-16
fix: use expo-build-properties to create the config plugin

Titozzz:
Instead of manually recreating everything, I've just used expo-build-properties.

This fixes that plugin so that it does not override other manifest queries from other plugins but gracefully merges everything.

# Overview

# Test Plan

### DIFF
--- a/plugin/src/index.ts
+++ b/plugin/src/index.ts
@@ -1,93 +1,45 @@
-import {
-  withAndroidManifest,
-  createRunOncePlugin,
-  ExportedConfigWithProps,
-  ExportedConfig,
-} from '@expo/config-plugins';
+import { ExportedConfig } from '@expo/config-plugins';
+import { withBuildProperties } from 'expo-build-properties';
 
-// eslint-disable-next-line import/no-commonjs, @typescript-eslint/no-var-requires
-const pkg = require('../../package.json');
-
-/**
- * @type {import('./types').ManifestQueries}
- * what we are trying to add:
- * <queries>
-    <package android:name="com.facebook.katana"/>
-    <package android:name="com.instagram.android"/>
-    <package android:name="com.twitter.android"/>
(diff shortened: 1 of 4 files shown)


plugin/src/index.ts
```diff
@@ -1,93 +1,45 @@
...
+    {
+      ...config,
+      android: {
+        ...config.android,
+        ...(props.enableBase64ShareAndroid
+          ? {
+              permissions: [
+                ...(config.android?.permissions ?? []),
+                'android.permission.WRITE_EXTERNAL_STORAGE',
+              ],
```
anthlasserre, line 22: Should we remove duplicates in case `WRITE_EXTERNAL_STORAGE` is already passed as a permission?
```suggestion
              permissions: [...new Set([
                ...(config.android?.permissions ?? []),
                'android.permission.WRITE_EXTERNAL_STORAGE',
              ])],
```

anthlasserre (approved):

plugin/src/index.ts
Titozzz: We could. But I think expo should already be doing that when computing the end result anyway

mikehardy (commented):
😬  @Titozzz my sincere apologies but I just merged #1622 which was in the same area, I didn't see that this PR was outstanding. So there are conflicts in it now but it looks like your PR would just supercede the change from 1622? (that change was mostly handling null or typing irregularities)

hirbod:
Friends, can we get this one here merged? We just figured that react-native-share was overwriting instead of merging.

jblarriviere:
@hirbod @mikehardy PR is up to date, ready to merge when you have time (lint fail seems unrelated to our changes)

MateusAndrade (approved):

MateusAndrade:
I will skip the CI, as the iOS build is failing because of a lock issue. I will fix it with a separate PR. 🚀

PR #1622, opened 2024-12-29
fix(ios): handle for edge case in Expo plugin when LSApplicationQueriesSchemes is an object and props are undefined

### LINKED ISSUE #1623: IOS: Expo plugin prebuild fails occasionally when `props` parameter is undefined, or `LSApplicationQueriesSchemes` is an object
### Steps to reproduce
1. Create a new Expo project using `npx create-expo-app`
2. Install plugin using `expo install react-native-share`
3. Try prebuilding using `expo prebuild`

### Expected behavior
Tell us what should happen
Prebuild should complete with no errors

### Actual behavior
Tell us what happens instead
Prebuild is currently throwing an error due to LSApplicationQueriesSchemes coming through as an object (this may be an edge case, it doesn't seem like the expected type at least)

### Environment
- **React Native version**: 0.76.5
- **React Native platform + platform version**: n/a
- **Typescript version** (if using typescript): n/a

### react-native-share
**Version**: 12.0.3

https://github.com/duncanlutz/react-native-share-issue

duncanlutz:
# Overview

I ran into two edge cases when trying to utilize this package, specifically when using the Expo plugin. First, when running `expo prebuild` within my real project, I received an error `Cannot read properties of undefined (reading 'ios')`.

Screenshot 2024-12-29 at 2 55 45 PM

Upon investigation, the error appeared to be occurring because the `props` parameter was undefined.

To further diagnose I created a minimal repo, only to run into another error upon prebuild, `TypeError: config.ios.infoPlist.LSApplicationQueriesSchemes is not iterable`.

Screenshot 2024-12-29 at 2 55 26 PM

I added some logging just inside my node_modules folder and discovered that LSApplicationQueriesSchemes was being provided as an object in my case (I'm not sure why this is occurring exactly, it's listed as an array within the Info.plist file).

To fix this, I added a function which checks whether the Query Schemes are an array as expected and if they are not returns an empty array. I also added a fallback for `props.ios` in case the props parameter is undefined as in my case.

# Test Plan

After updating the code, I replaced my existing code within my node_modules folders in both packages, and the prebuild process executed as expected. These tests cover scenarios where:

* LSApplicationQuerySchemes is undefined
* LSApplicationQuerySchemes is an array of strings but props was undefined
* LSApplicationQuerySchemes is an object

which from what I can tell covers all preexisting scenarios, plus the new edge cases I discovered.

- Fixes #1623

### DIFF
--- a/plugin/src/index.ts
+++ b/plugin/src/index.ts
@@ -63,6 +63,15 @@ const withAndroidManifestService = (config: ExportedConfig, props: WithSocialSha
   });
 };
 
+/**
+* Handles for edge case when LSApplicationQueriesSchemes is an object or undefined.
+*/
+const getIOSQuerySchemes = (config: ExportedConfig): Array<string> => {
+  return Array.isArray(config.ios?.infoPlist?.LSApplicationQueriesSchemes)
+    ? config.ios.infoPlist.LSApplicationQueriesSchemes
+    : [];
+}
+
 const withInfoPlist = (config: ExportedConfig, props: WithSocialShareProps) => {
   return {
     ...config,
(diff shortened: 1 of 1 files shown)


mikehardy (approved):
LGTM - thank you!

PR #1631, opened 2025-02-04
fix: handle RN 0.77 breaking change for event listener unsubscribe

rorygilchrist:
# Overview
Since RN 0.77 removeEventListener no longer exists on BackHandler, causing type errors. Updated ShareSheet to use new method of removing the subscription. More information can be found here: https://reactnative.dev/blog/2025/01/21/version-0.77#android

image

# Test Plan
Verified within my own production app that it now builds and doesn't cause type errors

### DIFF
--- a/src/components/ShareSheet.tsx
+++ b/src/components/ShareSheet.tsx
@@ -34,9 +34,9 @@ const ShareSheet: React.FC<React.PropsWithChildren<ShareSheetProps>> = ({
   }, [visible, onCancel]);
 
   React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


MateusAndrade (approved):

mikehardy:
CircleCI had some transient failure I saw in other repos as well from cocoapods not getting boost pod downloaded correctly. Rerunning it

MateusAndrade:
I will skip the iOS build and merge so we can fix this issue, and open a PR fixing the build after.

Issue #1624, opened 2025-01-02
Sharing video to Instagram Stories does not work on iOS


valerykonst:
### Steps to reproduce

```
     try {
            const shareOptions = {
                social: Share.Social.INSTAGRAM_STORIES, 
                appId:  'FB_APP_ID' // 'your_fb_app_id',
                backgroundVideo: videouri // video local uri 
            };
            Share.shareSingle(shareOptions); 
        } catch (error) {
            console.error('Error sharing to Instagram', error);
        }     
        
```
p.s. if I share image to stories (I pass backgroundImage), it works fine. 

### Expected behavior
Instagram should open share story

### Actual behavior
On iOS Instagram does not open (works on  Android). 

### Environment
- **React Native version**: 0.74.5
- **React Native platform + platform version**: iOS 18.1.1
- **Typescript version** (if using typescript): ~5.3.3

### react-native-share
**Version**: ^12.0.3

### Link to repo (highly encouraged)
https://github.com/valerykonst/Motivation

grettchen:
https://www.facebook.com/share/1bEGD4UabQ/?mibextid=QHi97vXeaPhvBiRL

Issue #1613, opened 2024-12-10
cannot open URL


lcaparada:
I am trying to share an image to Instagram Stories, but whenever I try to open it, the error "cannot open URL" occurs.
I don't know if the issue is with the app ID I'm passing or if it's a problem with the library.

 ```
 async function handleCaptureAndShareToInstagram() {
    try {
      const uri = await captureRef(viewRef, {
        format: "jpg",
        quality: 1,
      });

      console.log(uri);

      if (hasInstagramInstalled) {
        const { sharingCode } = await handleCreateRecipeSharingCode();

        await Share.shareSingle({
          stickerImage: uri,
          appId: "<my-app-add>",
          backgroundTopColor: "#ffffff",
          social: Social.InstagramStories,
          backgroundBottomColor: "#ffffff",
        });
      } else {
        await Share.open({ url: uri });
      }
    } catch (error) {
      console.error(error);
    }
  }
```
  
  -----------------------------------------
  
```
 <Container ref={reference}>
      <Card width={width}>
        <PhotoContainer>
          {photo && <StyledImage source={{ uri: photo }} />}
        </PhotoContainer>
        <TitleText>{name}</TitleText>
        <SubtitleText>{homeName}</SubtitleText>
        <Row>
          <IconImage
            source={require("../../../../../assets/icon.png")}
            onLoadEnd={() => setShareRecipeImageIsLoaded(true)}
          />
          <IconText>{iconName}</IconText>
        </Row>
      </Card>
    </Container>
```
 

 "react-native-view-shot": "^4.0.3",
 "react-native": "0.73.6",
 "react-native-share": "^12.0.3"

Podlipny:
Have you found any solution to this, I have the exact same problem for you? I am also using `react-native-view-shot` to capture images, I convert it to base64 and when I try to share stories, I get an error `cannot open URL` and get redirected to Apple Store.

```TS
  const shareInstagramStory = useCallback(async () => {
    try {
      const storyUri = await storyRef.current.capture()

      const shareOptions = {
        title: 'Share image to instastory',
        backgroundImage: await getBase64(storyUri),
        backgroundBottomColor: '#fefefe',
        backgroundTopColor: '#906df4',
        social: Share.Social.INSTAGRAM_STORIES,
        appId: '<my app key>',
      }

      // TODO: probably TS error in react-native-share, it throws error that Share.Social.INSTAGRAM
      // is not valid type of InstagramStoriesShareSingleOptions
[...]

PR #1633, opened 2025-02-09
fix: Instagram iOS sharing of stories

### LINKED ISSUE #1619: iOS: The app you shared from doesn’t currently support sharing to Stories
### Steps to reproduce
1. Install react-native-share and configure it as per the documentation.
2. Use the following code to attempt sharing to Instagram Stories:

```
import Share from 'react-native-share';

const shareToInstagramStories = async () => {
        try {
          const shareOptions = { 
              backgroundImage: 'http://urlto.png',
              appId: 'YOUR_FACEBOOK_APP_ID', // Replace with your actual App ID
              backgroundBottomColor: '#fefefe',
              backgroundTopColor: '#906df4',
              attributionURL: 'https://link.com',
              stickerImage: backgroundUri || ''
          };

    await Share.shareSingle({
      ...shareOptions,
      social: Share.Social.INSTAGRAM_STORIES,
    });
    console.log('Shared successfully!');
  } catch (error) {
    console.error('Error:', error.message);
  }
};
```
 

### Expected behavior 
The app opens Instagram Stories directly.
The selected image appears as the background image in the Instagram story editor.
A link tag (e.g., attribution URL) appears so the user can tap it.
The user can finalize and share the story directly from Instagram.

### Actual behavior
The app redirects directly to Instagram.
An error appears without opening the story dialog:
"The app you shared from doesn’t currently support sharing to Stories."
Image_20241217_152830

### Environment
- **React Native version**: ^0.71.8  
- **React Native platform + platform version**:  
   - **iOS**: [13.0]  
   - **Android**: 1.0.67 
   
### react-native-share
- **Version**: 12.0.3  
- **npm version or "master"**: 9.8.1

voslartomas:
# Overview

We had problems with sharing instagram stories  on iOS, first problem was with function always returning `cannot open URL`, but this seems not necessary as in app usually you check for instagram installation before sharing (at least that is our case).

First issue solved sharing images, but there was another problem with sharing videos, based on https://developers.facebook.com/docs/instagram-platform/sharing-to-stories/ it seems that videoBackground expects NSData, so I just simplified whole code a bit to convert to NSData as it is done with images.

This should close #1619 #1613  #1624

# Test Plan

I have tested those changes in our app and it works both for images and videos, even with sticker with link.

https://github.com/user-attachments/assets/112ce0b4-2070-4d91-bc6f-cd5dc52f0631


MateusAndrade:
@voslartomas can you check if the issue persists with last release? If that's the case, could we fix the conflicts with the PR? :)

voslartomas:
> @voslartomas can you check if the issue persists with last release? If that's the case, could we fix the conflicts with the PR? :)

Just tested newest release, it still has problem with not being able to open scheme url `[instagram-stories/share](instagram-stories://share?source_application=)`, but after removing that check part, it seems working fine. So removing that part which works now and keeping removal of scheme check.

MateusAndrade (approved):

younes0:
@MateusAndrade @voslartomas does this solve #1624 ?

mateustome:
@voslartomas @Daviddavidov98 same question here. How to add the custom link to show in instagram story?

Deveshb15:
> @voslartomas @Daviddavidov98 same question here. How to add the custom link to show in instagram story?

hey @mateustome were you able to figure this?

PR #1667, opened 2025-05-07
Fix Instagram singleShare iOS crash

dawidzawada:
# Overview
When I was calling `singleShare` like that, I was getting a crash:
```typescript
const shareOptions: ShareSingleOptions = {
  url: `data:image/png;base64,${base64}`
  social: Social.Instagram,
  type: 'image/*',
}

await Share.shareSingle(shareOptions)
```
I had a look at InstagramShare implementation and I've discovered that calling `performChanges` from `PHPhotoLibrary` requires following permissions:
```xml
<key>NSPhotoLibraryUsageDescription</key>
<string>This app requires access to the photo library to save and share images on Instagram.</string>
<key>NSPhotoLibraryAddUsageDescription</key>
<string>This app requires access to the photo library to save and share images on Instagram.</string>
```
I think It's not described enough in documentation, If someone is missing this properties he is getting a crash, if they were added but the user denies access, the error from the promise was the following:
```
 (NOBRIDGE) LOG  [Error: error]
```
After my changes it will be:
```
 (NOBRIDGE) LOG  [Error: Photo library access denied by user]
```
or:
```
 (NOBRIDGE) LOG  [Error: Photo library access not authorized]
```

Now, the promise will be rejected with proper message. I'm curious on your thoughts about adding a separate function handling these permissions this would remove a need for using libraries like `react-native-permissions` or `expo-media-library`. I'd be happy to implement that.


MateusAndrade (approved):
Thanks! 🚀

PR #1670, opened 2025-06-07
feat(ios): twitter/x shareSingle

akinncar:
# Overview
Add twitter/x shareSingle to iOS.

Closes: https://github.com/react-native-share/react-native-share/issues/1648

# Test Plan
```jsx
const shareToTwitter = async () => {
  const shareOptions = {
    message: 'Example Twitter',
    url: 'https://google.com',
    social: Share.Social.TWITTER,
  };

  try {
    const ShareResponse = await Share.shareSingle(shareOptions);
  } catch (error) {
  }
};
```

I didn't check if it's possible to share images, probably yes, it can be added later if necessary.

### DIFF
--- a/website/docs/share-single.mdx
+++ b/website/docs/share-single.mdx
@@ -94,6 +94,7 @@ You can pass the option that will be handled by the native code, similar to `Sha
 | **LINKEDIN**          |   ✅    | 🚫  | 🚫      |
 | **TELEGRAM**          |   ✅    | ✅  | 🚫      |
 | **VIBER**             |   ✅    | ✅  | 🚫      |
+| **TWITTER/X**         |   🚫    | ✅  | 🚫      |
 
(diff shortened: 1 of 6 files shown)


website/docs/share-single.mdx
```diff
@@ -94,6 +94,7 @@ You can pass the option that will be handled by the native code, similar to `Sha
 | **LINKEDIN**          |   ✅    | 🚫  | 🚫      |
 | **TELEGRAM**          |   ✅    | ✅  | 🚫      |
 | **VIBER**             |   ✅    | ✅  | 🚫      |
+| **TWITTER/X**         |   🚫    | ✅  | 🚫      |
```
akinncar, line 97: seems that we have the file for android but I didn't test yet, not sure if it's working or if anyone is using

Issue #1647, opened 2025-02-28
Nothing happens when share to telegram


floodsterdev:
### Steps to reproduce
Share text or image to telegram and whatsapp

### Expected behavior
Telegram/Whatsapp app should be opened

### Actual behavior

---Telegram

nothing happens
response => {  success: false, message: undefined }
...
BUG IN CLIENT OF UIKIT: The caller of UIApplication.openURL(_:) needs to migrate to the non-deprecated UIApplication.open(_:options:completionHandler:). Force returning false (NO).

---Whatsapp

TRY OPEN whatsapp
Try open view
WhastApp installed
Something went wrong
'Error sharing:', { [Error: Something went wrong]
  code: 'com.rnshare',
  nativeStackIOS: 
   [ '0   MyApp.debug.dylib                   0x000000010bea37dc RCTJSErrorFromCodeMessageAndNSError + 100',
     '1   MyApp.debug.dylib                   0x000000010c1ebe40 ___ZZN8facebook5react15ObjCTurboModule13createPromiseERNS_3jsi7RuntimeENSt3__112basic_stringIcNS5_11char_traitsIcEENS5_9allocatorIcEEEEU13block_pointerFvU13block_pointerFvP11objc_objectEU13block_pointerFvP8NSStringSH_P7NSErrorEEENK3$_0clES4_RKNS2_5ValueEPSQ_m_block_invoke.109 + 344',
     '2   MyApp.debug.dylib                   0x000000010bd9a88c -[WhatsAppShare handleError:code:rejectFn:] + 368',
     '3   MyApp.debug.dylib                   0x000000010bd99d10 -[WhatsAppShare tryToSendImage:resolve:reject:] + 956',
     '4   MyApp.debug.dylib                   0x000000010bd998dc -[WhatsAppShare shareSingle:reject:resolve:] + 336',
     '5   MyApp.debug.dylib                   0x000000010bd92180 -[RNShare shareSingle:resolve:reject:] + 1348',
     '6   CoreFoundation                      0x000000019e158e34 0013A8B1-2524-3534-B5BA-681AAF18C798 + 196148',
     '7   CoreFoundation                      0x000000019e157e7c 0013A8B1-2524-3534-B5BA-681AAF18C798 + 192124',
     '8   CoreFoundation                      0x000000019e1bba38 0013A8B1-2524-3534-B5BA-681AAF18C798 + 600632',
[...]

PR #1698, opened 2025-11-08
fix(ios) Viber & Telegram sharing for iOS 18+

JacobFJ:
# Overview

This is a **continuation and fix** for improvements introduced in #1588, which were inadvertently **not included from release `v11.0.4`**.

Resolve: #1647

# Test Plan

1. This change could not be tested on a real iOS device because the repository does not include the required provisioning profiles.

2. A working Expo example app would greatly help with thorough real-device testing (Telegram, WhatsApp, etc.). I currently don’t have the time to create one myself, so I’m hoping someone from the community can contribute it.

### DIFF
--- a/ios/ViberShare.m
+++ b/ios/ViberShare.m
@@ -25,14 +25,14 @@
     NSURL * shareURL = [NSURL URLWithString:urlViber];
     
     if ([[UIApplication sharedApplication] canOpenURL: shareURL]) {
-        [[UIApplication sharedApplication] openURL: shareURL];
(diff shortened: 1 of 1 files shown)


mikehardy:
> This change could not be tested on a real iOS device because the repository does not include the required provisioning profiles.

I think you can just open the sample app in Xcode and build it to your local developer device without much fuss? Possibly with a local change to use your personal developer profile temporarily for testing? Could be wrong, just saw that testing callout and thought I'd remembered doing something like what I suggest in the past...

PR #1700, opened 2025-11-26
feat(android): add useInternalStorage option for base64 sharing

### LINKED ISSUE #1695: Android WRITE_EXTERNAL_STORAGE Permission in Docs
https://react-native-share.github.io/react-native-share/docs/share-open#sharing-a-base64-file-format

According to the docs you need to add the `WRITE_EXTERNAL_STORAGE` permission in order to share a base64 encoded file on Android 

```xml
<!-- required for react-native-share base64 sharing -->
<uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
```

https://developer.android.com/reference/android/Manifest.permission#WRITE_EXTERNAL_STORAGE

**However for apps targeting API 30 or higher this permission does nothing**. And to get your app into the play store you need to at least target APi 34, so for any play store app this permission does nothing. 

This may be in part related to #1687 

I recently upgraded to version 12.2.0 from 10.2.1 and noticed that base64 image sharing was broken and came across the docs linked above saying that I should be using `WRITE_EXTERNAL_STORAGE`. I was surprised to see that the app that I was working on *did in fact* declare this permission even though it does nothing and needs to be removed. 

I was able to resolve my issue to share a base 64 encoded image on android after upgrading to 12.2.0 by setting `useInternalStorage: true` on my call to `Share.open`, but think that ultimately the documentation for this project should **not** be telling people to declare `WRITE_EXTERNAL_STORAGE` in 2025

realstone2:
- Introduced `useInternalStorage` option in ShareOptions and BaseShareSingleOptions to support internal storage caching for base64 file sharing on Android.
- Updated documentation to emphasize the necessity of this option for apps targeting Android API 30 and above.

Fixes #1695

# Overview

# Test Plan

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -64,6 +64,8 @@ interface BaseShareSingleOptions {
   recipient?: string;
   social: Exclude<Social, Social.FacebookStories | Social.InstagramStories>;
   forceDialog?: boolean;
+  /** Android Only: Store the temporary file in the internal storage cache */
+  useInternalStorage?: boolean;
 }
 
 interface BaseSocialStoriesShareSingleOptions extends Omit<BaseShareSingleOptions, 'social'> {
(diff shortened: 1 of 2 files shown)


MateusAndrade (approved):
Nice, thanks! 🚀

PR #1766, opened 2026-07-12
fix: add internal files path to Android FileProvider

### LINKED ISSUE #1764: Uri Permissions not set on Android
File sharing on Android seems to not set the correct permissions for URI sharing. Adding it manually makes it work again.

### Steps to reproduce
1. Clone the following repo: https://github.com/Belllene/rn-share-issue
2. Build and run the application
3. Press the button
4. Try to share with an app that can handle images, e.g. Google Photos

### Expected behavior
The image should be shared with the selected app.

### Actual behavior
The operation gets canceled.

If you copy to following values into the `Android` folder, it works.
In AndroidManifest.xml:
```xml

...
  <activity ...>
    ...
    <provider
      android:name="androidx.core.content.FileProvider"
      android:authorities="${applicationId}.rnshare.fileprovider"
      android:exported="false"
      android:grantUriPermissions="true">
      <meta-data
        android:name="android.support.FILE_PROVIDER_PATHS"
        android:resource="@xml/file_paths" />
    </provider>
```

Create a new file `android/app/src/main/res/xml/file_paths.xml` with the following content:
```xml
<paths>
    <files-path
        name="internal_files"
        path="." />
</paths>
```

### Environment
- **React Native version**: 0.83.6
- **React Native platform + platform version**: Android SDK 36
- **Typescript version** (if using typescript): 5.9.2

### react-native-share
**Version**: 12.3.1

https://github.com/Belllene/rn-share-issue

wanxiankai:
# Overview

Closes #1764.

The Android FileProvider bundled with `react-native-share` currently exposes downloads and cache files, but it does not expose files from the app's internal files directory. When a user shares a `file://` URI that points under internal files, `FileProvider.getUriForFile(...)` cannot match it against the configured paths, so receiving apps may fail to access the shared image/file.

This adds a `<files-path>` entry to the bundled `share_download_paths.xml`, matching the workaround described in the issue while keeping the existing provider authority and cache/download paths unchanged.

# Test Plan

Commands and integration checks run locally:

- Parsed `android/src/main/res/xml/share_download_paths.xml` with Python's XML parser: passed
- `git diff --check`: passed
- Reproduced the integration path with Expo SDK 55 and React Native 0.83.6 after applying this PR's XML change to `react-native-share` 12.3.1
- Confirmed Expo prebuild does not copy the dependency Manifest or XML into `android/app/src`; this is expected because React Native autolinks the package as an Android library
- Ran Android Gradle Plugin 8.12 `:app:processDebugMainManifest :app:mergeDebugResources`: passed
- Confirmed the merged application Manifest contains `cl.json.RNShareFileProvider`, the resolved `<applicationId>.rnshare.fileprovider` authority, and `@xml/share_download_paths`
- Confirmed the packaged library resource contains `<files-path name="rnshare3" path="." />`

These checks validate the Manifest/resource integration discussed in the follow-up. Device-level sharing still needs confirmation in the reproduction app.

### DIFF
--- a/android/src/main/res/xml/share_download_paths.xml
+++ b/android/src/main/res/xml/share_download_paths.xml
@@ -2,4 +2,5 @@
 <paths xmlns:android="http://schemas.android.com/apk/res/android">
     <external-path name="rnshare1" path="Download/" />
     <cache-path name="rnshare2" path="/" />
+    <files-path name="rnshare3" path="." />
 </paths>


Belllene:
Thanks for taking a stab at my issue! I added an additional comment to the issue. Based on the current behaviour of the module, I don't think this PR would fix it because the `share_download_paths.xml` is not placed in the Android folder by the prebuild.

wanxiankai:
Thanks for the follow-up. I verified this specific integration path against the Expo SDK 55 / React Native 0.83.6 reproduction setup.

Expo prebuild does not copy dependency manifests or resources into the app's `android/app/src` directory, but that is expected: autolinking includes `react-native-share/android` as an Android library, and Android Gradle Plugin merges the library manifest and packaged resources during the build.

After applying this PR's XML change to `react-native-share` 12.3.1, I ran AGP 8.12's `:app:processDebugMainManifest :app:mergeDebugResources` targets. The resulting application manifest contains `cl.json.RNShareFileProvider`, the resolved `<applicationId>.rnshare.fileprovider` authority, and the `@xml/share_download_paths` reference. The packaged library resource also contains `<files-path name="rnshare3" path="." />`.

So the files not appearing in the generated Android source directory should not prevent this fix from working. Could you test the PR branch in the reproduction app and confirm the device-level share flow? If it still cancels, the Android log around `RNSharePathUtil::compatUriFromFile` would help identify the remaining runtime issue.

Belllene:
Thank you @wanxiankai for your detailed explanation! I played around with some other modules yesterday and their manifests got merged in during pre-build. It seems there are different mechanisms to achieve the same goal? I'm still fairly new to react-native/expo.

Your change does indeed fix the reported issue. I changed the file in the `node_modules` folder the same way you did in this PR and file sharing worked as expected. Thank you for your work :)

Belllene:
Is anything missing for this to get merged?

